PR #1, opened 2025-07-10
[blackjack - Step1] Submit sebasnadu's mission

SebasNadu:
# kotlin-blackjack

Here is my Step1 PR, thanks in advanced to taking the time for this review, I really appreciate it😊.

## Questions

* I use abstract class for inherited Participant to Player and Dealer, Do you think that an open class on Player could simplify the code? What do you recommend me? I have also problems try to find when to use Interface, abstract and open classes, and more when In Kotlin you can implement also methods inside Interfaces.
* Do you think that having a `GameResult` property of each `Participant` is clean, or it's better separate this logic from participants?

### Model

* Introduced an abstract `Participant` class to encapsulate shared logic between `Player` and `Dealer`.
* Created a `Player` class extending `Participant`.
* Created a `Dealer` class extending `Participant`, responsible for managing the `Deck` and evaluating outcomes against each player.
* Implemented a `Card` class using `enum` types for `Suit` and `Rank`.
* Added a `Deck` class that initializes a standard shuffled 52-card deck.
* Added a `Hand` class representing each participant’s collection of cards and related logic.
* Added a `GameResult` property to `Participant` for tracking win/loss/tie results.

### Controller

* Implemented `BlackJackGame` to control the game flow and orchestrate interactions between model and view.

### View

* Handles user input and validation.
* Outputs game state and results to the console.


README.md
```diff
@@ -1 +1,58 @@
...
+### Model
+- [x] **Create Participant Class**
+  - [x] Should contain a name, hand
+  - [x] Should Add a card to the hand.
+  - [x] Should check for blackjack and busts
+  - [x] The name should be not empty.
+- [x] **Create Player Class inherit from a participant**
+- [x] **Create Dealer Class inherit from a participant**
+  - [x] Should contain a Deck and manage it.
+  - [ ] Should be able to automatically add cards to the hand.
```
Wordbe: If you've completed the code, please check the checkbox!

README.md
```diff
@@ -1 +1,58 @@
...
+    - [x] Should be comma separated names
+    - [x] Should contain a minimum of one player.
+    - [x] The names should be unique.
+- [x] OutputView
+  - [x] Should print the list of all participants and their hands.
+  - [x] Should print the final hands and score.
+  - [x] Should print the Final Result.
+  - [x] Add empty lines between each step on the game.
+
+### Todo
```
Wordbe: Here as well, please check the checkbox once the code is complete 🙂

src/main/kotlin/blackjack/controller/BlackJackGame.kt
```diff
@@ -0,0 +1,84 @@
+package blackjack.controller
+
+import blackjack.model.Dealer
+import blackjack.model.Player
+import blackjack.view.InputView
+import blackjack.view.OutputView
+import java.lang.Exception
+import kotlin.system.exitProcess
+
+class BlackJackGame {
```
Wordbe: Rather than having BlackJackGame hold state, it would be more stable to use such values only within the necessary functions and pass them as function arguments when needed. 😊

- dealer, players

src/main/kotlin/blackjack/controller/BlackJackGame.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    fun calculateResults() {
+        OutputView.printFinalHands(players, dealer)
+
+        players.forEach { dealer.setResultFor(it) }
+
+        OutputView.printResults(players, dealer)
+    }
+
+    private fun <T> retryUntilSuccess(block: () -> T): T {
```
Wordbe: How about testing this function as well?
If it's a private function, consider how it can be tested?

Give it a try!

src/main/kotlin/blackjack/controller/BlackJackGame.kt
```diff
@@ -0,0 +1,84 @@
...
+            players.forEach { it.addCard(dealer.dealCard()) }
+        }
+        OutputView.printParticipantsHands(players, dealer)
+    }
+
+    fun dealingPlayersCards() {
+        for (player in players) {
+            while (!player.hasBlackJack() && !player.isBusts()) {
+                val answer = retryUntilSuccess { InputView.getAnswer(player.name) }
+                if (answer == "y") {
```
Wordbe: Try creating a meaningful constant for this value. It will make your code more readable!

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,71 @@
...
+            }
+
+            else -> {
+                player.setWin()
+                this.setLose()
+            }
+        }
+    }
+
+    override fun getResults(): String {
```
Wordbe: While using view-related representations inside a domain object may be convenient during development, it's something that requires consideration.
A domain object can become dependent on the view, violating the Single Responsibility Principle.
For example, if something in the view needs to change, you might have to modify the domain object.
It would be good to explore alternative approaches as well. 🙂

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,71 @@
+package blackjack.model
+
+class Dealer(name: String = "Dealer") : Participant(name) {
+    val deck = Deck()
```
Wordbe: Let’s move them into the primary constructor.
Also, please add tests related to these variables. 🙂

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,32 @@
+package blackjack.model
+
+class Deck() {
+    private val _cards: MutableList<Card> = mutableListOf()
```
Wordbe: You used a backing property!
What was the reason for using it, and what are its advantages? 🤔

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    val cards: List<Card>
+        get() = _cards.toList()
+
+    init {
+        create()
+        shuffle()
```
Wordbe: The Deck object automatically calls the non-deterministic shuffle() function upon creation. Because of this, it’s difficult to perform deterministic tests, such as placing specific cards into the deck for testing purposes.
What kind of structure would be better for allowing more controlled testing? 🙂

src/main/kotlin/blackjack/model/GameResults.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.model
+
+class GameResults {
```
Wordbe: From the player's perspective, GameResult is counted only once, whereas from the dealer's perspective, it can be counted multiple times.
Although this isn't necessarily a problem, it's generally a good idea to separate concepts that have different characteristics.

Especially, it seems like each player doesn’t necessarily need to hold GameResult as a state.
Could we consider refactoring this a bit? 😊

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    fun getScore(): Int {
+        var aceCount = 0
+        var score = 0
+        dealtCards.forEach {
+            if (it.rank == Card.Rank.ACE) aceCount += 1
+
+            score += it.rank.value
+            if (score > BLACK_JACK && aceCount > 0) {
+                score -= 10
```
Wordbe: Shall we define these kinds of values as constants separately as well? 😊

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model
+
+class Hand {
+    val dealtCards: MutableList<Card> = mutableListOf()
```
Wordbe: The Hand object looks great! It seems you have a strong ability to abstract concepts. 😊

Wordbe: It would be easier to test if you define it in the primary constructor. 😊

Wordbe (changes requested):
Hello, Sebas!
Nice to meet you!

Thank you for your hard work in carefully implementing the Blackjack mission.

Please review the comments, reflect on them, and apply any necessary changes if possible.

---

Q1.
I use abstract class for inherited Participant to Player and Dealer, Do you think that an open class on Player could simplify the code? What do you recommend me?

It seems more semantically appropriate to inherit from the Participant! 

Q2. 
I have also problems try to find when to use Interface, abstract and open classes, and more when In Kotlin you can implement also methods inside Interfaces.

Try distinguishing between abstract classes and interfaces primarily based on whether they hold state or not. 

Q3.
Do you think that having a GameResult property of each Participant is clean, or it's better separate this logic from participants?

It would be great to determine which object is responsible for handling the game result! 

---

Thank you! 😊

src/main/kotlin/blackjack/controller/BlackJackGame.kt
SebasNadu: 🤔🤔...Thanks for the feedback! I see your point: keeping the controller stateless by passing dealer and players as arguments would make it more modular and testable. I'll revisit the structure and consider remove the state in the controller or separate the game state into a dedicated class, so the controller only coordinates actions without holding internal state.

Appreciate the insight, it definitely helps me think more clearly about responsibility boundaries😮.

src/main/kotlin/blackjack/model/Dealer.kt
SebasNadu: 🤔You're right, I did it to simplify the view during development, but I completely agree with your point. If the view changes, that change shouldn't ripple into the model. I’ll explore cleaner alternatives to keep responsibilities properly separated. Thanks for the insight!

src/main/kotlin/blackjack/model/Deck.kt
SebasNadu: Yeah, I used a backing property so I can modify the list inside the class but keep it hidden from outside. That way, no one can mess with the cards directly. It helps keep things safe and controlled. But I’m open if you think there’s a better way! 👀

src/main/kotlin/blackjack/model/Deck.kt
SebasNadu: Yes, you’re right. The first step would be to remove shuffle() from the init and call it from outside instead. Then, accept a list of cards in the primary constructor with an empty list as the default. That way, you can test shuffle() with custom cards. For creating a standard deck internally, maybe use a static method in the companion object that calls create() to generate the deck and returns the instance.

I like that! this approach lets you pass a specific list of cards to Deck, so you can test shuffle() by comparing the original list with the internal list after shuffling.  Nice! Thanks for the idea, I didn't find the way to test the shuffle() before!

src/main/kotlin/blackjack/model/GameResults.kt
SebasNadu: mmm yes, we discussed this with my teammate. Maybe the simple and cleanest way, it is to separate the result in other class or object, that holds or generate the results and extract also this from the participants.🤔 I'll think about this. Thank you!

src/main/kotlin/blackjack/model/Hand.kt
SebasNadu: Yes, good point. I'll create like ACE_ADJUSMENT or something similar

src/main/kotlin/blackjack/model/Deck.kt
SebasNadu: After revisiting the code, I noticed I don’t actually use cards outside, so technically the backing property isn’t needed right now, but I decided to keep it for future-proofing, in case I need controlled access later.
Open to alternatives if you have a better suggestion!

SebasNadu:
Hi Wordbe! 👋

I’ve implemented the changes and suggestions you shared, along with some additional improvements:
-  Introduced DealerResultTracker and PlayerResultTracker, both inheriting from the ResultTracker interface. This allows each participant to track results independently and flexibly.
- Added a GameJudge object, responsible for comparing the dealer’s and players’ hands and determining the game outcome.
- Simplified the controller by moving out unnecessary logic, making it more focused and maintainable.
- Improved variable naming and increased the use of constants to enhance readability and maintainability.
- Expanded test coverage, including internal functions instead of private ones to enable testing previously inaccessible logic.
- Enhanced error message formatting for better clarity in the console.
- Added colored card output in the view to improve the user experience.
- Restructured the model package into three main domains: card, participant, and result for better organization.

Let me know what you think! 😊

Example current execution
---
CleanShot 2025-07-12 at 23 21 53

src/main/kotlin/blackjack/model/card/Card.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model.card
+
+data class Card(val suit: Suit, val rank: Rank) {
```
Wordbe, line 3: It’s impressive how you improved the card display to make it look nicer in the console! 👍👍

However, I noticed one area that could be improved—the domain object like Card is currently dependent on the View. For instance, as you did, when updating the card UI, you had to modify the Card domain object. This indicates that the responsibilities between the domain and the view are not clearly separated.

It would be a good idea to decouple these concerns, so that changes in the view (UI) don’t require modifications to the domain logic. This will help maintain separation of concerns and make your codebase more flexible and maintainable.

Other domain objects, including the Card, also seem to need some changes!

src/main/kotlin/blackjack/controller/BlackJackGame.kt
Wordbe: You changed the method from private to internal to make it testable — it's great that you strengthened your test coverage!
However, this also exposed the method’s visibility in production code.

An alternative is to extract the private method into another object and make it public there.
This allows for testing while also separating responsibilities. Just something to keep in mind! 😊

src/main/kotlin/blackjack/model/Deck.kt
Wordbe: Yes, sounds good. Backing properties help protect internal property mutations from external access, but overusing them can hurt readability. It's best to use them only when truly necessary! 😊

src/main/kotlin/blackjack/model/GameJudge.kt
```diff
@@ -0,0 +1,31 @@
...
+    internal fun evaluate(
+        dealer: Dealer,
+        player: Player,
+    ) {
+        val result = dealer vs player
+        player.recordResult(result)
+        dealer.recordResult(result.inverse)
+    }
+
+    private infix fun Dealer.vs(player: Player): Result =
```
Wordbe: Using the infix vs method is a great idea! 😊 But instead of using an extension function, how about defining the method directly in the Participant class? (or Dealer)

src/main/kotlin/blackjack/model/participant/Participant.kt
```diff
@@ -0,0 +1,28 @@
+package blackjack.model.participant
+
+import blackjack.model.card.Card
+import blackjack.model.card.Hand
+import blackjack.model.result.Result
+import blackjack.model.result.ResultTracker
+
+abstract class Participant(val name: String, protected val resultTracker: ResultTracker) {
```
Wordbe, line 8: Does the Participant really need to hold the game result? I think it might be better if the GameJudge determines the outcome at the end and provides the result. Or do you have a different opinion? 🤔

src/test/kotlin/blackjack/model/PlayerTest.kt
```diff
@@ -0,0 +1,27 @@
...
+    fun `should throw an error if player name is empty`() {
+        assertThrows<IllegalArgumentException> { Player("") }
+    }
+
+    @Test
+    fun `should return an string with the name + dealt cards`() {
+        val player = Player("Sebas")
+        player.addCard(Card(Card.Suit.DIAMONDS, Card.Rank.ACE))
+        player.addCard(Card(Card.Suit.DIAMONDS, Card.Rank.ACE))
+        println(player)
```
Wordbe: How about changing it to an assert statement? 🙂

Wordbe (changes requested):
Hello Sebas!

Thank you for reflecting the code changes with such attention to detail and care. Your thoughtful implementation really shows the effort you put into crafting clean, well-structured code. It's evident that you take pride in your work, and it definitely shows in the quality of what you've delivered.

I've left a few more thoughts in the comments — please take a look and revise where possible.

Feel free to share your thoughts anytime. Thanks again for the excellent work! 🙌

src/main/kotlin/blackjack/model/card/Card.kt
SebasNadu: You’re right🤔. I’ll move the display logic out of the domain and delegate it to the view layer instead💪.

src/main/kotlin/blackjack/model/GameJudge.kt
SebasNadu: Yes, I was also unsure where the best place for this method would be. In the end, I left it inside GameJudge since that’s the only place it’s currently used, and its purpose felt specific to that context. But you’re right, defining it directly in the Dealer class would make the design cleaner and more intuitive.

I had similar questions myself, so I really appreciate you sharing your thoughts, it’s very helpful! 🙂

src/test/kotlin/blackjack/model/PlayerTest.kt
SebasNadu: Yes, totally. Sorry!😅

src/main/kotlin/blackjack/model/participant/Participant.kt
SebasNadu: That’s a great point, and I actually considered a similar idea after your first round of feedback. Initially, I also thought about keeping the result outside the Participant. But after reviewing the requirements for step2, I decided it’s more appropriate for the Participant to hold its result state.

In step2, the result is tightly coupled with the participant’s bet, the outcome directly determines their profit or loss. In that context, storing the result inside the participant feels natural, close to the real world and keeps all related information together.

Even in step1, although the link isn’t as strong, this approach avoids introducing an additional data structure just to associate a result with a participant. Since all the necessary information is already available within the participant, it made sense to store the result there. Conceptually too, the result feels like part of the participant’s state for me, so why not encapsulate it there?

That said, I did decouple the actual evaluation logic by extracting it into the GameJudge, so the dealer is no longer responsible for determining results. But I prefer to leave the result state inside the participants, avoiding extra data structure and prioritizing an easy and simple connection between the view and the full data from the Participant.

Thank you again for the thoughtful question, please let me know if you consider that this decision is problematic, I don't have also any problem to modify my app structure.

SebasNadu:
Hi Wordbe! 👋

Thanks again for your valuable feedback, it really helped me clarify some key concepts and improve my understanding of the project.

Here’s what I’ve changed:
- Moved AnsiColor to the view layer, where it belongs.
- Delegated more print logic to the view, especially for participants and their hands.
- Error messages now appear in red for better visibility.
- Created DealerExtensions.kt to centralize dealer-related extension functions like Dealer.vs().
- Kept the result state inside Participant for convenience and to support the logic needed in step 2.

Let me know what you think! 💪

src/main/kotlin/blackjack/model/participant/Participant.kt
Wordbe: You gave me a sufficient explanation that the presence of this state is justified by its relevance to the object, and I found it convincing. I hope you can make good use of this structure in the upcoming steps. Thank you for the detailed explanation. 😊

src/main/kotlin/blackjack/model/GameJudge.kt
```diff
@@ -0,0 +1,21 @@
...
+import blackjack.model.participant.Participants
+import blackjack.model.participant.Player
+import blackjack.model.participant.vs
+
+object GameJudge {
+    fun evaluateAll(participants: Participants) {
+        participants.players.forEach { evaluate(participants.dealer, it) }
+    }
+
+    internal fun evaluate(
```
Wordbe, line 13: Unless there's another specific reason, I’d prefer changing the access modifier more conservatively. 😊

src/main/kotlin/blackjack/model/card/Card.kt
```diff
@@ -0,0 +1,30 @@
...
```
Wordbe, line 3: What is the reason you used a data class? 🙂

src/main/kotlin/blackjack/model/result/DealerResultTracker.kt
```diff
@@ -0,0 +1,18 @@
...
+            Result.WIN to 0,
+            Result.LOSE to 0,
+            Result.TIE to 0,
+        )
+
+    override fun record(result: Result) {
+        resultMap[result] = resultMap.getValue(result) + 1
+    }
+
+    override fun toString(): String {
```
Wordbe, line 15: You're overriding the toString method of a domain object to handle view representation. This means that if the view's representation changes, the domain object will also need to be modified.
So, how about letting the view layer handle all of this instead?
(This applies to other objects as well.) 😊

Wordbe (approved):
Hello Sebas!

Thank you for putting deep thought into the reviews and doing your best to reflect the feedback.

I can see many improvements compared to the initial code, so I’ll approve and merge this PR now.

However, I’d appreciate it if you could consider and apply the comments I left in the next step.

Great work, and thank you! 😊

src/main/kotlin/blackjack/model/card/Card.kt
SebasNadu: Good point. I initially defined it as a data class because, in the example, the first two cards seemed to be sorted. I thought using data would let me call .sorted() easily for that. But then I forgot about it 😅. Thanks for pointing it out, I’ll double-check, and if I’m not going to actually using it, I’ll remove it 👍.

SebasNadu: ok, I was confuse. I thought that data card also override the compareTo() but not. Thank you!

PR #2, opened 2025-07-10
[BlackJack - Step1] Submit Bojana's Mission

bojana-petroska:
### BlackJack Step 1

This PR implements the BlackJack game, structured into the following components: `Model`, `Controller`, `InputView`, and `OutputView`.

---

### Model

- **Participant** (abstract class): Contains shared logic between `Player` and `Dealer`.
- **Player**: Is inherited from `Participant`.
- **Dealer**: Is inherited from `Participant` and adds:
  - Deck management.
  - Comparing game results with each player.
- **Card**: Represents a single card, using `Suit` and `Rank` enums.
- **Deck**: Holds and shuffles 52 standard cards.
- **Hand**: Belongs to each `Participant` and manages their cards.
- **GameResult**: Tracks each participant’s outcome.

---

### Controller

- **BlackJackGame**: Coordinates the game flow and connects model and view layers.

---

### View

- **InputView**: Handles and validates user input.
- **OutputView**: Displays game output to the user.

---

### Feedback Requested

We’d appreciate your thoughts on the following:

1. **Abstraction Design**: We used an abstract class (`Participant`) for shared logic between `Player` and `Dealer`.  
   - Do you think this is a good use of an abstract class?  
   - Would using an interface have been a better design choice in this context?

2. **Single Responsibility Principle**:  
   - Does each class follows well the Single Responsibility Principle?  
   - Are there any responsibilities that feel overloaded or misplaced?

3. **GameResult Management**:  
   - Does storing game results inside each `Participant` make sense from a design perspective, or would a centralized game statistics class would be better?

---

Thanks so much for your time and feedback!  
— *Bojana*


README.md
```diff
@@ -1 +1,58 @@
-# kotlin-blackjack
\ No newline at end of file
+# kotlin-blackjack
+
+## Features
+
+### Model
```
da-nyee, line 5: Well documented 👍

README.md
```diff
@@ -1 +1,58 @@
...
+  - [x] Should print the list of all participants and their hands.
+  - [x] Should print the final hands and score.
+  - [x] Should print the Final Result.
+  - [x] Add empty lines between each step on the game.
+
+### Todo
+  - [ ] Add customize exception messages
+  - [ ] Add constants to Input and Output Prompts
+  - [ ] Print cards with color (black and red)
+  - [ ] Update the hasBlackJack method to return true only in the first hand
```
da-nyee, line 58: Are you planning to handle these later?

src/main/kotlin/blackjack/controller/BlackJackGame.kt
```diff
@@ -0,0 +1,84 @@
...
+import blackjack.model.Dealer
+import blackjack.model.Player
+import blackjack.view.InputView
+import blackjack.view.OutputView
+import java.lang.Exception
+import kotlin.system.exitProcess
+
+class BlackJackGame {
+    val dealer = Dealer()
+    var players = listOf<Player>()
```
da-nyee: How about wrapping this collection?

src/main/kotlin/blackjack/controller/BlackJackGame.kt
```diff
@@ -0,0 +1,84 @@
...
+
+    fun start() {
+        createPlayers()
+        dealFirstCards()
+        dealingPlayersCards()
+        dealingDealersCards()
+        calculateResults()
```
da-nyee: Nice method extraction — it’s clear what each method is doing 👍

da-nyee: However, consider making them `private`, since they’re only used within this class.

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,30 @@
+package blackjack.model
+
+abstract class Participant(val name: String) {
```
da-nyee, line 3: Yes! I think using an `abstract class` is better than an `interface` for now.
If you need a higher level of abstraction later, you can consider converting it to an `interface`.
I’d recommend approaching abstraction step-by-step — from a concrete `class`, to an `abstract class`, and then to an `interface`.

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model
+
+class Hand {
```
da-nyee, line 3: Good name choice 👍

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -0,0 +1,11 @@
+package blackjack.model
+
+class Player(name: String) : Participant(name) {
+    override fun getResults(): String {
+        return when {
+            gameResults.wins > 0 -> "$name: Win"
+            gameResults.loses > 0 -> "$name: Lose"
+            else -> "$name: Tie"
```
da-nyee: I guess the model now depends on the view — how about moving this logic to the `OutputView` instead?
`Model` should stay independent from `Controller` and `View`.

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    fun setResultFor(player: Player) {
+        val playerScore = player.getScore()
+        val dealerScore = this.getScore()
+        when {
+            player.isBusts() -> {
+                player.setLose()
+                this.setWin()
+            }
```
da-nyee, line 28: In Kotlin, there’s a useful function called `apply` that you can use when setting values to properties.
It helps make your code more concise and improves readability by grouping related property assignments together.

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,71 @@
...
+                player.setWin()
+                this.setLose()
+            }
+        }
+    }
+
+    override fun getResults(): String {
+        val winText = if (gameResults.wins > 0) "${gameResults.wins} Win" else ""
+        val loseText = if (gameResults.loses > 0) "${gameResults.loses} Lose" else ""
+        val tieText = if (gameResults.ties > 0) "${gameResults.ties} Tie" else ""
```
da-nyee: I'd recommend moving this logic to the view as well, as I mentioned.

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,29 @@
+package blackjack.view
+
+object InputView {
+    fun getPlayersNames(): List<String> {
+        println("Enter the names of the players (coma-separated):")
+        val input = readLine() ?: ""
```
da-nyee:
```suggestion
        val input = readlnOrNull() ?: ""
```

Consider using this method to prevent a `NullPointerException`, which is one of the most problematic runtime exceptions.

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,29 @@
...
+        }
+        require(playersNames.toSet().size == playersNames.size) {
+            "Names should be unique."
+        }
+        return playersNames
+    }
+
+    fun getAnswer(playerName: String): String {
+        println("Would $playerName like to draw another card? (y for yes, n for no)")
+        val input = readLine()?.trim()?.lowercase() ?: ""
```
da-nyee: Here too.

src/test/kotlin/blackjack/model/PlayerTest.kt
```diff
@@ -0,0 +1,25 @@
+package blackjack.model
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+
+class PlayerTest {
```
da-nyee, line 7: I'd suggest adding more tests, such as `DealerTest`, `HandTest`, and so on.

da-nyee (changes requested):
Hi Bojana! 🙌
You did a really good job — nice work 👍 
It seems like you have a solid understanding of OOP and how to apply it in your code.
I left some comments — feel free to check them out and let me know if you have any questions 👀

README.md
bojana-petroska: yes, we wanted to have a to do list and hope we will manage it given the time constraints!

src/main/kotlin/blackjack/model/Participant.kt
bojana-petroska: Thanks! This is a very useful hint, I have spent the whole week wondering about in which case an abstract class is beneficial and in which an interface : )

src/main/kotlin/blackjack/controller/BlackJackGame.kt
bojana-petroska: Thank you for the suggestion, it really makes sense. I will implement a new Collection for Players.

src/main/kotlin/blackjack/view/InputView.kt
bojana-petroska: good suggestion, I actually used it in Lotto : )))

src/main/kotlin/blackjack/model/Player.kt
bojana-petroska: Yes, makes sense. I did it for both, the Player and the Dealer.

src/main/kotlin/blackjack/model/Dealer.kt
bojana-petroska: do you mean like this:
```
player.isBusts() -> {
     player.apply { setLose() }
     this.apply { setWin() }
}
```

bojana-petroska: I used it in the tests for some of the chained card additions

da-nyee: @bojana-petroska 

```kotlin
player.isBusts() -> apply {
    player.setLose()
    this.setWin()
}
```

I meant like this !

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,90 @@
...
+    ) {
+        println("$dealer - Total: ${dealer.getScore()}")
+        players.forEach { println("$it - Total: ${it.getScore()}") }
+    }
+
+    fun printPlayersResults(players: List<Player>) {
+        players.forEach { player ->
+            val results = when {
+                player.gameResults.wins > 0 -> "${player.name} Win"
+                player.gameResults.loses > 0 -> "${player.name}: Lose"
```
da-nyee, line 62: You moved them 👍

src/test/kotlin/blackjack/model/HandTest.kt
```diff
@@ -0,0 +1,50 @@
+package blackjack.model
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class HandTest {
```
da-nyee, line 6: 👍

da-nyee (approved):
Hi Bojana! 🙌
Thank you for applying the feedback 🙇‍♀️

I went through all the changes, and I think it's good to merge this now, since we have one more step to go.
I also left a comment — feel free to take a look. Let me know if you have any other questions.

Cheers, and good luck with the next step 💪

PR #3, opened 2025-07-10
[blackjack step 1] submit pebencze's mission

pebencze:
Hello!

Thank you for reviewing my mission! :)

This project is about simulating a blackjack card game respecting the official rules. I have tried using inheritance, lambda functions, function chaining, data classes, extension functions and single responsibility.
I would appreciate some feedback on how I implemented these and if they respect object calisthenics.

I have noted some questions, as well:
Q1: How should we name players inside the Players class?
Q2: Should we use multiple decks?
Q3: State in Controller (even if private) is bad practice?

Wishing you a good weekend ahead and thank you once again.

Best,

Petra


README.md
```diff
@@ -1 +1,93 @@
...
+- [x] display cards
+- [x] display sum
+
+- [x] calculate results
+  - [x] player 
+  - [x] dealer 
+- [x] display results
+
+## Considerations
+Q1: How should we name players inside the Players class? </br>
```
da-nyee, line 91: I think `player` works well, though `people` or something similar could work too — just a matter of preference.

README.md
```diff
@@ -1 +1,93 @@
...
+Q2: Should we use multiple decks? </br>
```
da-nyee, line 92: You could use it, but I wouldn't recommend it, since using a single deck is sufficient for now.
Doing this might be over-engineering for this task.

README.md
```diff
@@ -1 +1,93 @@
...
+Q3: State in Controller (even if private) is bad practice?</br>
```
da-nyee, line 93: Yes, this is considered a bad practice.
`Controller` and `View` shouldn’t hold any state — only `Model` should manage state.

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,98 @@
...
+            OutputView.displayDealerDrawMessage()
+            this.drawCard(deck)
+        }
+    }
+
+    private fun Players.dealCards() {
+        players.forEach { dealCards(it) }
+    }
+
+    fun dealCards(player: Player) {
```
da-nyee, line 55: Consider making it `private`.

da-nyee: Because this method is only used here.

src/main/kotlin/blackjack/model/CardDeck.kt
```diff
@@ -0,0 +1,29 @@
+package blackjack.model
+
+class CardDeck(
+    val deck: ArrayDeque<Card> = ArrayDeque(),
```
da-nyee, line 4: 👍

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,23 @@
+package blackjack.model
+
+abstract class Participant {
```
da-nyee, line 3: Is there a specific reason for using an `abstract class` rather than an `interface`?

src/main/kotlin/blackjack/model/ResultEvaluator.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun toString(): String {
+        return "Dealer: $wins wins, $losses losses, $draws draws"
+    }
+}
+
+data class PlayerResult(var win: Boolean = false, var draw: Boolean = false, val name: String = "") {
+    override fun toString(): String {
+        if (win) return "$name: win"
+        if (draw) return "$name: draw"
+        return "$name: loss"
```
da-nyee: It seems like the model depends on the view.
To follow the Single Responsibility Principle, consider moving this logic into the view.

src/main/kotlin/blackjack/model/ResultEvaluator.kt
```diff
@@ -0,0 +1,66 @@
...
+    }
+}
+
+class ResultEvaluator(val players: Players, val dealer: Dealer) {
+
+    fun calculateResults(): Pair<DealerResult, List<PlayerResult>> {
+        return Pair(calculateDealerResults(), calculateAllPlayersResults())
```
da-nyee: What about wrapping the response as well, instead of using a `Pair`?

src/main/kotlin/blackjack/model/HandCards.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class HandCards(val cards: MutableList<Card> = mutableListOf()) {
+    val total: Int
+        get() = calculateTotal()
+
+    fun add(card: Card) {
+        cards.add(card)
+    }
+
+    fun calculateTotal(): Int {
```
da-nyee: I'd recommend making this method `private`, since it's only used here.

src/main/kotlin/blackjack/model/HandCards.kt
```diff
@@ -0,0 +1,37 @@
...
+        var total = 0
+        for (card in cards) {
+            total += card.rank.value
+        }
+        total -= discount(total)
+        return total
+    }
+
+    fun discount(total: Int): Int {
```
da-nyee: This one too.

src/main/kotlin/blackjack/model/HandCards.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model
+
+class HandCards(val cards: MutableList<Card> = mutableListOf()) {
```
da-nyee, line 3: How about adding tests for this class?

da-nyee (changes requested):
Hi Pebencze! 🙌
I think you have great potential to grow, since you've asked thoughtful questions — I answered all of them.
I also left a few comments, so feel free to take a look and let me know if you have any other questions 👀

README.md
pebencze: Thank you!:) I specifically mean lines like this: `dealerResult.losses = players.players.count() { it.isNotBusted() }
`

README.md
pebencze: I will try to refactor. :)

src/main/kotlin/blackjack/model/Participant.kt
pebencze: Well, our consideration was that Player and Dealer ARE a Person, and do not HAVE a Person. They also only inherit from an abstract class and not from multiple interfaces. Other than that, I will need to do my research. 🤓

src/main/kotlin/blackjack/model/ResultEvaluator.kt
pebencze: Fair! Would be more readable.

src/main/kotlin/blackjack/model/HandCards.kt
pebencze: We are actually testing crucial features of it in TotalTest, but maybe I could rename the test file...

README.md
pebencze: The reviewer of my teammate told that this code section would be fine for this project's scope. In my previous Lotto project I have tried to refactor my Controller and it makes the code more readable, but also more bloated because of passing parameters. Do you think I should refactor?

pebencze:
Hello Dani!

Thank you for your valuable feedback! I have tried to implement most feedback except for two, where I have left comments. :) I would appreciate if you could answer those and review my new implementation. 

Also, I have learnt about the State Design Pattern, which is super cool, but if I were to implement that, I would need to discard my whole codebase for step 2. Also, I am already starting a new project next week. What do you advise?
A. Shall I just continue with my current codebase and implement the new features (betting money)?
B. Discard, try State model.
C. Continue from my current base, but experiment with the States in a separate package/repository/branch.

Have a great weekend!

Best,

Petra

da-nyee:
Hi @pebencze !

> Also, I have learnt about the State Design Pattern, which is super cool, but if I were to implement that, I would need to discard my whole codebase for step 2. Also, I am already starting a new project next week. What do you advise?
A. Shall I just continue with my current codebase and implement the new features (betting money)?
B. Discard, try State model.
C. Continue from my current base, but experiment with the States in a separate package/repository/branch.

Well, I'd recommend giving the State Pattern a try and going with `B`. `A` seems more like just completing the mission, which might feel comfortable but could limit your growth. `C` feels like doing double the work, so it doesn’t seem very efficient.

Choosing `B` might take more time since you’ll have to rewrite some code, but it could broaden your perspective and boost your confidence by helping you develop a new coding skill.

Completing the mission is important — but learning something meaningful along the way matters even more.

That said, this is just my personal opinion — feel free to do what feels best for you.

src/main/kotlin/blackjack/model/HandCards.kt
da-nyee: Oh, I completely missed that point — thank you for catching it.

README.md
da-nyee: In that case, how about sending a message to `Players` instead of calling it directly from `ResultEvaluator`?
It might make the code a bit cleaner and more object-oriented.

README.md
da-nyee: This mission is more like a small project, so it might not be a big deal that the `Controller` holds state.
However, I'd still recommend refactoring it, since this is practice for building much larger projects.
It's a good opportunity to explore different ways of writing code.

src/main/kotlin/blackjack/model/ResultEvaluator.kt
```diff
@@ -0,0 +1,56 @@
+package blackjack.model
+
+data class DealerResult(var wins: Int = 0, var losses: Int = 0, var draws: Int = 0)
+
+data class PlayerResult(var win: Boolean = false, var draw: Boolean = false, val name: String = "")
```
da-nyee, line 5: Wrapping them 👍

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    fun dealerResultString(result: DealerResult): String {
+        return "Dealer: ${result.wins} wins, ${result.losses} losses, ${result.draws} draws"
+    }
+
+    fun playerResultToString(result: PlayerResult): String {
+        if (result.win) return "${result.name}: win"
+        if (result.draw) return "${result.name}: draw"
+        return "${result.name}: loss"
```
da-nyee, line 45: You moved the code 👍

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,53 @@
...
+            println()
+        print("${participant.name}'s cards: ${participant.cardsToString()}")
+        println(" – Total: ${participant.handCards.total}")
+    }
+
+    fun dealerResultString(result: DealerResult): String {
+        return "Dealer: ${result.wins} wins, ${result.losses} losses, ${result.draws} draws"
+    }
+
+    fun playerResultToString(result: PlayerResult): String {
```
da-nyee, line 42: How about making them `private`?

da-nyee (approved):
Hi pebencze! 🙌
I went through the changes and left a few more comments — feel free to check them out.
I think it's okay to merge now, but I'd recommend addressing the new comments in the next step.
Looking forward to your next PR! 🤸‍♀️

PR #4, opened 2025-07-10
[Blackjack - Step1] Submit san-ghun's mission

san-ghun:
# Blackjack Step 1

## Overview

This PR is to implement requirements of the Blackjack project step1. The project is a simulation of card game Blackjack in terminal text input and output. The step1 requirements are for a simplified version of Blackjack. In this game, the player or dealer with a total closest to 21 without going over wins.

## Main Features

- Implement and design under application of MVC pattern.
   - Models: [ Card, Dealer, Hand, Player, Rank, Stats ] + [ Playable (interface) ]
   - Views: [ InputView, OutputView ]
   - Controllers: [ GameManager, CardManager, PlayerManager, DealerManager, StatsManager ]
   - Utils: [ CardGenerator ]

- `GameManager` is an object and a main controller unit among the controllers to run program and other managers working under `GameManager`.
- `CardManager` take one input parameter `CardGenerator` which can be extracted into inheritable object in future and can be used as DI(Dependency Injection).
- interface `Playable` inherit to `Player` and `Dealer` to share role as playable in game.
- `Stats` calculates and stores final results of players and dealer.

## Additional Notes

In this project, I tried to apply MVC pattern under my understanding and it was mostly focused on how to separate responsibility of a main controller. That is why I have several managers as controller units. Also wanted to follow class design to limit access to other class and object. TDD is still hard behavior to apply during the project but my brain is getting used to it. I am really glad that I am getting better than previous project in many aspect.

Hope I get a lot of feedback from your review, so I can learn and grow further more.

Thank you!


src/main/kotlin/blackjack/Application.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack
+
+import blackjack.controller.GameManager
+
+fun main() {
+    val gm = GameManager
```
Wordbe: I think it does not need a variable!

src/main/kotlin/blackjack/controller/CardManager.kt
```diff
@@ -0,0 +1,46 @@
+package blackjack.controller
+
+import blackjack.model.Card
+import blackjack.utils.CardGenerator
+
+class CardManager(generator: CardGenerator = CardGenerator) {
```
Wordbe: Object instances usually don’t need to receive dependencies through their constructors. It’s good to think about why that is! 🤔

src/main/kotlin/blackjack/model/Card.kt
```diff
@@ -0,0 +1,6 @@
+package blackjack.model
+
+data class Card(val name: String) {
+    val digit = name.substring(0, name.length - 1)
```
Wordbe: Parsing strings can easily cause issues if the format changes, and it’s also hard for people to read. Instead, try creating objects with a defined structure!

src/main/kotlin/blackjack/controller/GameManager.kt
```diff
@@ -0,0 +1,51 @@
+package blackjack.controller
+
+import blackjack.view.InputView
+import blackjack.view.OutputView
+
+object GameManager {
```
Wordbe: I think using `~Manager` might be overkill. Especially if you just access something like dealManager.dealer directly, it defeats the purpose of having a Manager! 🙂

src/main/kotlin/blackjack/controller/GameManager.kt
```diff
@@ -0,0 +1,51 @@
...
+    val cardManager = CardManager()
+    val playerManager = PlayerManager()
+    val dealerManager = DealerManager()
+    val statsManager = StatsManager(playerManager.players, dealerManager.dealer)
+    val inputView = InputView
```
Wordbe: There’s no need to assign an object to a variable. Especially since some parts use this variable while others call static methods, mixing the two ways.

src/main/kotlin/blackjack/controller/PlayerManager.kt
```diff
@@ -0,0 +1,48 @@
+package blackjack.controller
+
+import blackjack.model.Card
+import blackjack.model.Player
+
+class PlayerManager {
+    private var _players: List<Player> = emptyList()
```
Wordbe: Why did you use backing property? What benefits does it have? 🙂

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,22 @@
+package blackjack.model
+
+data class Hand(val cards: List<Card> = emptyList()) {
```
Wordbe: Why did you use data class? 🙂

Wordbe: Let's define the numbers in this class as meaningful constants to improve readability.

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,22 @@
...
+        _hand = Hand(deque.toList())
+    }
+
+    override fun calculateHand(): Int {
+        return _hand.calculateCards()
+    }
+
+    fun shouldDrawCardOrNot(): Boolean {
+        val score = calculateHand()
+        return score <= 16
```
Wordbe: Shall we define the magic numbers as constants?

src/main/kotlin/blackjack/model/Rank.kt
```diff
@@ -0,0 +1,25 @@
...
+    TEN("10", 10),
+    JACK("J", 10),
+    QUEEN("Q", 10),
+    KING("K", 10),
+    NONE("0", 0),
+    ;
+
+    companion object {
+        fun of(digit: String): Rank {
+            return Rank.entries.firstOrNull { it.digit == digit } ?: NONE
```
Wordbe: How about throwing an exception instead of using NONE?

src/main/kotlin/blackjack/model/Stats.kt
```diff
@@ -0,0 +1,48 @@
+package blackjack.model
+
+data class Stats(val players: List<Player>, val dealer: Dealer) {
+    private var _playerBoard = initPlayerBoard()
```
Wordbe: Instead of defining internal variables, try returning the result from a method.
Going forward, when creating state in an object, do so only when it's truly necessary. 💡

src/main/kotlin/blackjack/model/Stats.kt
```diff
@@ -0,0 +1,48 @@
...
+        val loseCount = playerBoard.values.count { it == WIN }
+        val tieCount = playerBoard.values.count { it == TIE }
+        stats["win"] = winCount
+        stats["lose"] = loseCount
+        stats["tie"] = tieCount
+        _dealerStats = stats.toMap()
+    }
+
+    companion object {
+        const val LOSE = 0
```
Wordbe: Try defining it using enums or descriptive strings instead of raw numbers. It will improve the readability!

src/main/kotlin/blackjack/utils/CardGenerator.kt
```diff
@@ -0,0 +1,33 @@
...
+        val values = mutableListOf<String>()
+        digits.map { first ->
+            figures.map { second ->
+                values += "$first$second"
+            }
+        }
+        return values.toList()
+    }
+
+    fun generateCard(suit: String): Card {
```
Wordbe: This seems unnecessary. Try removing it and simplifying the code. 😊

Wordbe (changes requested):
Hello, San!
Nice to meet you! 😊

Thank you for your hard work writing the Blackjack mission code with care.
Please review the feedback and apply any necessary changes after considering them!

Here are some general areas that stood out for improvement:

Use backing properties only when truly necessary. They can make the code harder to read and understand. Instead of using numbers to represent states, consider using enums. They make the code much more readable. Avoid creating and parsing strings unnecessarily. Manage data in structured formats from the beginning—it’ll be easier to maintain. Strive to simplify your code as much as possible. Simple code is always easier to manage than complex logic. Replace magic numbers with constants. This improves clarity, reduces mistakes, and enhances overall understanding. 

Thank you again! Keep up the great work! 🙌

src/main/kotlin/blackjack/Application.kt
san-ghun: Agree! Thanks for pointing out!

src/main/kotlin/blackjack/controller/CardManager.kt
san-ghun: Thank you for a topic to consider. 
What I can answer is, in this case, I guess receiving one-use only dependency(`CardGenerator`) is an overkill(?) for the `CardManager` instance.

san-ghun: I am considering to redesign `Card` class to have feature of generating `Cards` that only hold 52 instances of `Card` and only return pointer of the instance, preventing to create new `Card` instance any occasion.

src/main/kotlin/blackjack/model/Card.kt
san-ghun: Yes, I agree. I was also skeptical to assign parsed string into variable. I reimplement "Card" class with two enum class, I hope resubmitted implementation is better with you review.

src/main/kotlin/blackjack/controller/GameManager.kt
san-ghun: Yes, I agree on lack of efficiency on 'DealerManager'.
But is it okay to separate controllers into several to controller dedicated models? It is my first try separating a main controller into many. I am not sure, is it a good approach to separate responsibility of the main controller?

src/main/kotlin/blackjack/controller/GameManager.kt
san-ghun: Thanks! I also thought about the needs for assigning object into variable. It is good to know from your review.

src/main/kotlin/blackjack/controller/PlayerManager.kt
san-ghun: I used backing property because I wanted to not expose `players` to be editable. I want to expose `players` read-only and only able to add or remove by `PlayerManager`.

src/main/kotlin/blackjack/model/Hand.kt
san-ghun: Well, I though I am going to override some data class method to manipulate `Hand` for output view with `toString()` or addition method. But eventually it didn't go the way I have thought. I am making it into normal class.

src/main/kotlin/blackjack/model/Dealer.kt
san-ghun: Yes! I am on it! Thanks!!

src/main/kotlin/blackjack/model/Hand.kt
san-ghun: Yes! Thanks for letting me know gracefully!

src/main/kotlin/blackjack/model/Rank.kt
san-ghun: Yes! Good point! It is so much better!

src/main/kotlin/blackjack/model/Stats.kt
san-ghun: Yes, agree! I am introducing enums for this and in `OutputView` object. Thank you!

src/main/kotlin/blackjack/utils/CardGenerator.kt
san-ghun: Yes, I removed the `CardGenerator` object including the methods.

src/main/kotlin/blackjack/model/Stats.kt
san-ghun: Yes, now I know what is difference between calculated property and instance property(or field). 
I remove backing property for `playerBoard` and replace it into calculated property.
Thanks for pointing out!

san-ghun:
Hey, Suno!

Thanks for your thoughtful review! I have managed to apply review you mentioned. A main change since your last review is that I have reimplemented `Card` class into `PlayingCard` class with enum classes `Rank` and `Suit`. Because of the reimplementation of "PlayingCard" class, I edited lines of code. Sorry for adding confusion by reimplementing of class and hope you understand. 

# Main Changes

## Replacement of `Card` class to `PlayingCard` class

**Problem**: There are high possibility of creating for 'Card' instances those holding same values, and might provide confusion of managing 'Card' instances.

**Solution**
- redesign card class, removing 'Card' class and adding new class 'PlayingCard' with companion object that holding only 52 instances of 'PlayingCard'
- move card generation logic from 'CardGenerator' object into 'PlayingCard' class
- remove 'CardGenerator' object
- add 'Suit' enum class to create 'PlayingCard' instance using with 'Rank' enum class

## Remove `DealerManager` class

**Problem**: 'DealerManager' has no functionality over direct access to 'Dealer' instance.

**Solution**
- remove 'DealerManager' and give 'GameMaster' a direct access to 'Dealer' instance

## Introduction of `Result` enum class to represent 'result' of game

**Problem**: Previously 'Stats' class and 'OutputView' used number to represent result of game for players and dealer. It is having duplication of same code and usage of numbers to represent result of game provide lack of readability.

**Solution**
- introduce new enum class 'Result' to represent the result of game for players and dealer

## Remove magic numbers

**Problem**: There are too many magic numbers in codebase.

**Solution**
- introduce constants for magic numbers inside of `companion object`

---

I have additional questions related with the code review using GitHub.

1. When can or should "resolve" 'conversation' or 'feedback' or 'review comment'? 
    - Should I let it remain unresolved, or can I resolve after I applying feedback in new commit?
2. What would be your recommendation to have a good manner of code review, as reviewee? 
    - I am not sure leaving comments of message like what I did is good or bad.
3. What does "reviewer" expect from "reviewee"? 
    - I think with this answer, I can provide more and precise context to my future reviews.

Thanks you!

src/main/kotlin/blackjack/controller/PlayerManager.kt
Wordbe: Great! But since it may be hard to read, please use it only where it's truly necessary. 😊

src/main/kotlin/blackjack/controller/CardManager.kt
Wordbe: Great Thank you!

src/main/kotlin/blackjack/controller/GameManager.kt
Wordbe: As I mentioned, the current structure feels a bit excessive.
Also, rather than naming everything as "Manager", it would be better to give names that reflect the domain's role—so they act as living objects with clear responsibilities. 🙂

src/main/kotlin/blackjack/model/Hand.kt
Wordbe: Good Thank you!

src/main/kotlin/blackjack/controller/CardManager.kt
```diff
@@ -0,0 +1,47 @@
...
+    private var _cardMap: Map<PlayingCard, Boolean> = emptyMap()
+    val cardMap: Map<PlayingCard, Boolean>
+        get() = _cardMap
+
+    init {
+        cards = PlayingCard.Deck.toList().shuffled()
+        _cardMap = initiateCardMap()
+    }
+
+    private fun initiateCardMap(): Map<PlayingCard, Boolean> {
```
Wordbe: Why is a Map needed here? What does the Boolean in the Map represent?
How could this be made clearer?
Is it possible to implement this without using a Map?

src/main/kotlin/blackjack/controller/PlayerManager.kt
```diff
@@ -0,0 +1,42 @@
...
+    val players: List<Player> get() = _players
+
+    fun addPlayer(name: String) {
+        val mutableList = _players.toMutableList()
+        val player = Player(name)
+        mutableList.add(player)
+        _players = mutableList.toList()
+    }
+
+    // TODO: resolve indent depth
```
Wordbe: Remove the TODO and simplify the code as much as possible! ✂️

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,28 @@
...
+            score -= ACE_REDUCE_VALUE
+            aceCounts--
+        }
+        if (score < ACE_REDUCE_VALUE && values.contains(BIG_ACE_VALUE)) {
+            score += ACE_REDUCE_VALUE
+        }
+        return score
+    }
+
+    fun toText(): String {
```
Wordbe: If it's something needed in the View layer, how about handling it in the View instead of the Domain layer? 🙂

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,26 @@
+package blackjack.model
+
+data class Dealer(override val name: String = "Dealer") : Playable {
+    private var _hand = Hand()
```
Wordbe: Why is a backing property necessary here? 🤔

src/main/kotlin/blackjack/model/PlayingCard.kt
```diff
@@ -0,0 +1,28 @@
+package blackjack.model
+
+class PlayingCard private constructor(val rank: Rank, val suit: Suit) {
+    val value: Int get() = rank.value
+    val string: String get() = rank.digit + suit.form
```
Wordbe: Also, try separating parts that are only needed in the view layer here!

src/main/kotlin/blackjack/model/Result.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.model
+
+enum class Result(val value: Int) {
```
Wordbe: Why is value needed here? Could it possibly be removed? 🤔

src/main/kotlin/blackjack/model/Stats.kt
```diff
@@ -0,0 +1,45 @@
+package blackjack.model
+
+class Stats(val players: List<Player>, val dealer: Dealer) {
+    val playerBoard: Map<Player, Result> get() = calculatePlayerBoard()
```
Wordbe: Is playerBoard really necessary?
Try to minimize internal state as much as possible. 🙂

src/test/kotlin/blackjack/Fixture.kt
```diff
@@ -0,0 +1,20 @@
+package blackjack
+
+import blackjack.model.PlayingCard
+import blackjack.model.Rank
+import blackjack.model.Suit
+
+object Fixture {
```
Wordbe, line 7: Good 👍

Wordbe (changes requested):
Hello!

Thank you for writing the code so carefully and for reflecting on the review.

I’ve left some comments on parts that might need further consideration, so please take a look! 😊

---

Q1. When can or should "resolve" 'conversation' or 'feedback' or 'review comment'?Should I let it remain unresolved, or can I resolve after I applying feedback in new commit?

Yes, please fix the issues, leave a comment, and then resolve the conversation! 

Q2. What would be your recommendation to have a good manner of code review, as reviewee?

If you have a reason, feel free to share your opinion politely and confidently! 🙂

Q3. I am not sure leaving comments of message like what I did is good or bad.What does "reviewer" expect from "reviewee"?I think with this answer, I can provide more and precise context to my future reviews.

It’s really helpful if you share what you worked on—what was challenging and how you addressed it. That way, we can understand your thought process better. Or you can also suggest alternative approaches if you have any! 

Thank you!

src/main/kotlin/blackjack/controller/CardManager.kt
san-ghun: Yes, thanks for pointing out! I also cannot find any use of it. I will remove the Map.
I am also thinking to remove `CardManager` and introduce new model class like `Deck` for the functionality.

src/main/kotlin/blackjack/controller/PlayerManager.kt
san-ghun: Yes, I was looking away this part because it was difficult for me to improve this part. I try my best to simplify the code as much as possible! Thanks!

src/main/kotlin/blackjack/model/Hand.kt
san-ghun: Yes, you're right. I will move it to the View

src/main/kotlin/blackjack/model/Dealer.kt
san-ghun: I wanted to private the 'hand', but I think it doesn't have to, like you mentioned. I am removing it and alter to variable.

src/main/kotlin/blackjack/model/Result.kt
san-ghun: Yeah. Actually it has no use. lol
I am removing it.

src/main/kotlin/blackjack/model/Stats.kt
san-ghun: Okay, I will apply your suggestion. 
To remove `playerBoard` from `Stats`, I introduce new property 'result' to `Player` and `Dealer`. And I edit functions to calculate and update result of players and dealer.

san-ghun:
Hey, Suno!

Thanks for your review! There are two main changes since your last review. I replaced `CardManager` into `Deck`, and removed `PlayerManager`. These changes aim to reduce complexity of controllers and make the role and names more clear and enhance readability. There are several changes to apply review you mentioned.

# Main Changes

## Replace `CardManager` and introduce new class `Deck`

**Problem**: The name 'CardManager' is too ambiguous and not clear to represent the role.

**Solution**
- introduce new class `Deck` as an wrapper collection, and remain essential functionalities for 'Deck'.

## Remove `PlayManager` and move remained logic into `GameMaster`

**Problem**: The role and name of `PlayerManager` is ambiguous, and initiation of a player is complex

**Solution**
- remove `PlayerManager`
- let players and dealer to initiate with 'hand'

# Noticeable Changes

- move View related functionalities from `Hand` and `PlayingCard` to View layer
- remove a state property `playerBoard` from `Stats` and move the feature to `Playable` with `Result`
- remove unnecessary backing properties to enhance readability
- remove unnecessary states from `GameMaster`

---

Additionally, thanks for your answers from previous comment!
Thanks for your review, all the time!

src/main/kotlin/blackjack/controller/GameMaster.kt
```diff
@@ -0,0 +1,103 @@
...
+        OutputView.displayFinalState(players, dealer)
+        val winStatistics = calculateStatistics(players, dealer)
+        OutputView.displayFinalResults(winStatistics)
+    }
+
+    private fun initDealer(): Dealer {
+        val dealer = Dealer()
+        dealer.hand.initCards()
+        return dealer
+    }
```
Wordbe, line 28: It seems more appropriate for the `Dealer` to handle this. Try moving the code inside the `Dealer`. 🙂

src/main/kotlin/blackjack/controller/GameMaster.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    private fun initPlayers(): List<Player> {
+        val players = mutableListOf<Player>()
+        val names = InputView.retryable { InputView.readPlayerNames() }
+        names.forEach { name ->
+            val player = Player(name, Hand().initCards())
+            players.add(player)
+        }
+        return players.toList()
```
Wordbe, line 37: Instead of declaring variables beforehand, try returning a `List<Player>` in a functional style!

src/main/kotlin/blackjack/controller/GameMaster.kt
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    private fun calculateStatistics(
+        players: List<Player>,
+        dealer: Dealer,
+    ): Stats {
+        val winStatistics = Stats(players, dealer)
+        winStatistics.updateDealerStats()
+        return winStatistics
+    }
```
Wordbe, line 102: Try assigning the role to Stats!

src/main/kotlin/blackjack/controller/GameMaster.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    private fun drawAndDisplayAndReturnFalse(
+        player: Player,
+        receiveCard: () -> PlayingCard,
+    ): Boolean {
+        player.drawCard(receiveCard())
+        OutputView.displayCurrentHand(player)
+        return false
+    }
```
Wordbe, line 85: The receiveCard function goes as deep as 3 levels, which makes it hard to read. It would be great to improve this.

src/main/kotlin/blackjack/controller/CardManager.kt
Wordbe: Great. It seems a nice object. Thank you.

src/main/kotlin/blackjack/model/Stats.kt
```diff
@@ -0,0 +1,33 @@
+package blackjack.model
+
+class Stats(val players: List<Player>, val dealer: Dealer) {
+    var dealerStats = mapOf(Result.WIN to 0, Result.LOSE to 0, Result.TIE to 0)
+
+    fun updateDealerStats() {
```
Wordbe, line 6: It would be better to remove the state and create it within the method to return it. This could be a tradeoff between performance and state management. But if it's a lightweight calculation, I would prefer to eliminate state management. 🙂

Wordbe (approved):
Hello San,

Thank you for thoughtfully considering the blackjack mission review and carefully applying the feedback.
I will approve and merge this PR here! Great work.

Please review the additional comments I left and apply them when you start the next step.
Thank you. 🙏

src/main/kotlin/blackjack/controller/GameMaster.kt
san-ghun: Totally make sense!

src/main/kotlin/blackjack/controller/GameMaster.kt
san-ghun: Yes! I also can see how to make it better with functional style! I will try!

src/main/kotlin/blackjack/controller/GameMaster.kt
san-ghun: I don't fully understand what you are saying but I will try on step2!

src/main/kotlin/blackjack/controller/GameMaster.kt
san-ghun: Yes, it is. I am going to remove 'receiveCard'.

src/main/kotlin/blackjack/model/Stats.kt
san-ghun: Okay, I may have to consider about this that with the view between performance and state management. I will keep that in mind! Thanks!

PR #5, opened 2025-07-10
[Blackjack - Step1] Submit Deniz' mission

deniz-oezdemir:
### PR Ready for Review

Hi there,

I hope you are doing well!

This PR is ready for review. I'd appreciate your thoughts on a few questions that came up while working on it:

1. **Naming inside `Players` class**
   Currently, we refer to the list of players as `players`. Would something else be a clearer for this collection?

2. **Number of decks**
   Should we simulate multiple decks, like in a real casino? Right now, it's a single deck. Using more might better reflect how Blackjack is usually played.
   Anyways, like in most casinos we have capped the maximum amount of players at 7 so we will not run out of cards with one deck.

3. **Controller state**
   We keep some private variables in the `Controller` to track state. Is this an acceptable tradeoff, or should we use a more declarative approaches and pass variables between functions instead?

4. **`toString()` in `ResultsEvaluator` data classes**
   We’ve overloaded `toString()` to make it easier to print results. Should we instead move formatting into a separate function inside `OutputView`?

Additionally, if you spot anything that could be written in a more Kotlin-idiomatic way, I’d really value that feedback.

Thanks in advance for your time. All tips and suggestions are much appreciated.

Best,
Deniz


README.md
```diff
@@ -1 +1,93 @@
-# kotlin-blackjack
\ No newline at end of file
+# kotlin-blackjack
+Implementation of the Blackjack card game in Kotlin.
```
pkeugine, line 2: Nice work on organizing README and stacking commits 👍 

It'll be much better if we could add titles to some commits that don't have it,
and name them so that it reflects what that commits contain.
For example, there are two commits named `test(Dealer): test drawing logic for dealer`, but only one of them contains the actual test.

Screenshot 2025-07-11 at 11 48 58 AM

src/test/kotlin/study/DSLTest.kt
```diff
@@ -0,0 +1,34 @@
...
+ *          hard("Kotlin")
+ *      }
+ *      language {
+ *          "Korean" level 5
+ *          "English" level 3
+ *      }
+ * }
+ */
+
+class DSLTest {
```
pkeugine, line 9: This test file is not allowing me to run all the tests at once.
I think you've tried to remove this file in`.gitignore`, but it slipped into this PR.
Please remove this file. :)

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
+package blackjack.model
+
+class Dealer(
+    override val name: String = "Dealer",
+    override var handCards: HandCards = HandCards(),
+) : Participant() {
+
+    fun shouldDraw(): Boolean {
+        return handCards.total <= 16
```
pkeugine: I think we can give a name to the number 16, and treat it as a magic number.
Same goes to number 21, since it's a special number in blackjack.
How does this sound?

src/main/kotlin/blackjack/model/Players.kt
```diff
@@ -0,0 +1,18 @@
+package blackjack.model
+
+class Players(val players: List<Player>) {
```
pkeugine, line 3: **Q. Currently, we refer to the list of players as players. Would something else be a clearer for this collection?**

What were the other options that you and your pair thought of?
I'd like to know that first. :)

src/main/kotlin/blackjack/model/CardDeck.kt
```diff
@@ -0,0 +1,29 @@
+package blackjack.model
+
+class CardDeck(
```
pkeugine, line 3: **Q. Should we simulate multiple decks, like in a real casino? Right now, it's a single deck. Using more might better reflect how Blackjack is usually played.**

Since this decision is purely about the rule and flow about your blackjack implementation,
it's totally up to you to decide.

Personally, I think it'll be helpful for you to upgrade the game by using multiple decks,
since you'll be able to experience 'updates in project planning'. 👍

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,98 @@
...
+import blackjack.model.Players
+import blackjack.model.ResultEvaluator
+import blackjack.view.InputView
+import blackjack.view.OutputView
+import blackjack.view.OutputView.displayInitialCardsMessage
+
+class Controller {
+    private val deck = CardDeck()
+    private var _players: Players? = null
+    private val dealer = Dealer()
```
pkeugine: **Q. We keep some private variables in the Controller to track state. Is this an acceptable tradeoff, or should we use a more declarative approaches and pass variables between functions instead?**

Personally, I think it's fine in this level of project scale.

However, I want to know what that `tradeoff` is that you've thought of!
What are the pros and cons of this?

src/main/kotlin/blackjack/model/ResultEvaluator.kt
```diff
@@ -0,0 +1,66 @@
...
+    override fun toString(): String {
+        return "Dealer: $wins wins, $losses losses, $draws draws"
+    }
+}
+
+data class PlayerResult(var win: Boolean = false, var draw: Boolean = false, val name: String = "") {
+    override fun toString(): String {
+        if (win) return "$name: win"
+        if (draw) return "$name: draw"
+        return "$name: loss"
```
pkeugine: **Q. We’ve overloaded toString() to make it easier to print results. Should we instead move formatting into a separate function inside OutputView?**

Yes, I strongly recommend moving this logic to `OutputView`, since it's tightly coupled to 'console based UI'. In different kind of view other than console, we might want to make our output different, hence putting each of those output logic in their own respectable classes.

As for `toString()`, it's mostly used for debugging or logging purposes since debugging and logging output don't change rather we use one kind of view or another.

src/main/kotlin/blackjack/model/ResultEvaluator.kt
```diff
@@ -0,0 +1,66 @@
...
+            return PlayerResult(name = player.name)
+        if (dealer.isBusted())
+            return PlayerResult(win = true, name = player.name)
+        return when {
+            player.handCards < dealer.handCards -> PlayerResult(name = player.name)
+            player.handCards > dealer.handCards -> PlayerResult(win = true, name = player.name)
+            else -> PlayerResult(draw = true, name = player.name)
+        }
+    }
+}
```
pkeugine: I noticed this red indicator at the end of some files.
It would be helpful to find out why having a new line at the end of the file is preferred.  
**Hint**: [it's related to missing empty lines](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file) 😉  

In IntelliJ, there's an option to automatically add a newline at the end of a file. It’s worth checking that out.

src/main/kotlin/blackjack/model/Suit.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model
+
+enum class Suit(val suit: String) {
+    HEARTS("♥"),
+    DIAMONDS("♦"),
+    CLUBS("♣"),
+    SPADES("♠"),
```
pkeugine: I see you've added a visual effect for this game 👍 
Just like my previous comment, I personally think this visual effect also belongs to `OutputView`,
since there's a chance for us to use different way of showing the cards in other output strategy.

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,31 @@
...
+        }
+        return input.split(",").map { it.trim() }
+    }
+
+    fun promptForDraw(player: Player): Boolean {
+        println("\nWould ${player.name} like to draw another card? (y for yes, n for no)")
+        val input = readln().trim()
+        if (input.isEmpty()) {
+            throw IllegalArgumentException(ErrorMessage.EMPTY_INPUT.toString())
+        }
```
pkeugine:
```suggestion
        require(input.isNotEmpty()) { ErrorMessage.EMPTY_INPUT.toString() }
```

We can utilize `require` for this logic!

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,31 @@
...
+        return when (input) {
+            YES -> true
+            NO -> false
+            else -> throw IllegalArgumentException(ErrorMessage.EMPTY_INPUT.toString())
+        }
+    }
+
+    private const val YES = "y"
+    private const val NO = "n"
```
pkeugine: These constants made me think :  
* (This is my personal opinion) Do they really require names? I think it's common for anybody to think 'y' input is a yes, and 'n' input is a no. Naming them actually made it longer in code.

* The placement of these constants could be in the top of the object, since that is the most common convention. Or to make it simpler, since they are only used in one method, keep it inside that method.

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,45 @@
...
+        println(players.players.joinToString())
+    }
+
+    fun displayInitialCardsMessage(players: Players) {
+        println("\nDealing two cards to dealer, ${players.players.joinToString()}.")
+    }
+
+    fun displayAllCardsMessage(participant: Participant) {
+        print(participant.name + "'s cards: ")
+        println(participant.cardsToString())
```
pkeugine:
```suggestion
    println("${participant.name}'s cards: ${participant.cardsToString()}")

```

I think we can achieve the same output with less `println` method call by refactoring like above.
Also, if we need to print multiple lines, we can use multi-line string feature that Kotlin provides.

Why don't we use them to refactor our view related logic?

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,98 @@
...
+            OutputView.displayAllCardsMessage(player)
+        }
+    }
+
+    fun wantsToDraw(player: Player): Boolean {
+        repeat(MAX_TRIES){
+            try {
+                return InputView.promptForDraw(player)
+            } catch (e: IllegalArgumentException) {
+                println(e.message)
```
pkeugine: Although this is a simple logic, I think view related classes should take care of printing out errors as well. Give them the exception, messages, causes, or anything, and let them take care of printing what they want. How does this sound?

pkeugine: If we extend this idea, repeating `MAX_TRIES` until we get the desired input can be done in view as well. Controller and domain related logics don't need to know how many tries were taken before we actually play blackjack. They just want to play the game with nice player names and their 'yes's and 'no's. :)

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,98 @@
...
+        drawInitialCards()
+        OutputView.displayFirstCardMessage(dealer)
+        players.forEach { OutputView.displayAllCardsMessage(it) }
+        players.dealCards()
+        dealer.dealCards()
+        displayCardsAndTotal()
+        printResults()
+    }
+
+    fun printResults() {
```
pkeugine:
```suggestion
    private fun printResults() {
```

Let's keep methods as private if they aren't used by other classes.
Please check other methods across the whole project as well.

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,98 @@
...
+    fun Dealer.dealCards() {
+        while (this.shouldDraw()) {
+            OutputView.displayDealerDrawMessage()
+            this.drawCard(deck)
+        }
+    }
+
+    fun Players.dealCards() {
+        players.forEach { dealCards(it) }
+    }
```
pkeugine: Keeping these extension functions in between controller methods will make maintaining the project very difficult, since we're extending domain's logic in controller. Why don't we move these to somewhere else?

I do understand this happened because we want to print the status in these methods, but we'll have to take out `OutputView` logic from these domain logic to properly implement MVC pattern.

src/test/kotlin/blackjack/DeckTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import blackjack.model.Card
+import blackjack.model.CardDeck
+import blackjack.model.Rank
+import blackjack.model.Suit
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.assertThrows
```
pkeugine: Let's not forget to check if there are unused imports after refactoring.

src/test/kotlin/blackjack/ResultsTest.kt
```diff
@@ -0,0 +1,124 @@
...
+    fun `test dealer is not busted, player is not busted, DEALER WINS`() {
+        val playersCards = mutableListOf(Card(Rank.ACE, Suit.DIAMONDS), Card(Rank.NINE, Suit.DIAMONDS)) // total 20
+        val dealersCards = mutableListOf(Card(Rank.ACE, Suit.DIAMONDS), Card(Rank.KING, Suit.DIAMONDS)) // total 21
+        val dealer = Dealer(handCards = HandCards(dealersCards))
+        val player = Player("test", HandCards(playersCards))
+        val players = Players(listOf(player))
+        val evaluator = ResultEvaluator(players, dealer)
+        val results = evaluator.calculateResults()
+        OutputView.displayAllCardsMessage(dealer)
+        OutputView.displayAllCardsMessage(players.players[0])
```
pkeugine: Adding in `OutputView` in `ResultEvaluator` related  test seems unnecessary unless we use it as a debugging purpose. If it was used for debugging purpose, we can remove it after debugging, so that it doesn't get included in the commit.

If we wanted to make an integration test, there must be some kind of an assertion that checks if the `OutputView` behaved like we wanted it to.

Oh, and I think the name of this test class can be changed to `ResultEvaluatorTest`, since all the tests are about asserting it's logic.

src/test/kotlin/blackjack/TotalTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        assertEquals(handCards.total, 17)
+        handCards.add(Card(Rank.QUEEN, Suit.CLUBS))
+        assertEquals(handCards.total, 17)
+    }
+
+    @Test
+    fun `test if many aces are reduced`(){
+        val handCards = HandCards()
+        handCards.add(Card(Rank.ACE, Suit.DIAMONDS))
+        assertEquals(handCards.total, 11)
```
pkeugine, line 35: I want to know when you use `assertThat` and `assertEquals`. What are the differences between them? :)

build.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     testImplementation("org.junit.jupiter", "junit-jupiter", "5.10.2")
     testImplementation("org.assertj", "assertj-core", "3.25.3")
     testImplementation("io.kotest", "kotest-runner-junit5", "5.8.0")
+    testImplementation(kotlin("test"))
```
pkeugine: What are the reasons behind the inclusion of this dependency?
Also, I want you to tell me what this `kotlin("test")` does. 🙂

pkeugine (changes requested):
Hello Deniz,
I am doing well, and I hope you're good as well.
Nice work on blackjack mission! The code was very easy to read 👍 

I left some comments so it'll be nice for you to check them out before we move onto the next step.

If you have any questions or opinions, feel free to leave comments in this PR or hit me with a DM.

Happy hacking! 🎴 🎴 🎴

README.md
deniz-oezdemir: Yes, one of those was by mistake. Good catch! :-)

src/main/kotlin/blackjack/model/Players.kt
deniz-oezdemir: We thought of two options:
(1) Rename the class to "Table"
(2) Rename the attribute to "seats" or "spots"

src/main/kotlin/blackjack/controller/Controller.kt
deniz-oezdemir: The tradeoff for us in using private variables was that it's
- more readable, and
- less code for passing variables between functions
vs.
- more difficult to write tests
- more difficult implementing concurrency if needed

src/main/kotlin/blackjack/view/InputView.kt
deniz-oezdemir: Thanks for the suggestion!

src/main/kotlin/blackjack/view/OutputView.kt
deniz-oezdemir: Yes, I will try to reduce the calls to `println`.

src/test/kotlin/blackjack/ResultsTest.kt
deniz-oezdemir: Yes, we will remove these method calls in the test cases.

Yes, we will rename the test class. :-)

src/test/kotlin/blackjack/TotalTest.kt
deniz-oezdemir: `assertThat(...).isEqual` is more readable and less prone to confusion of the `actual` and the `expected`. So generally we should probably use `assertThat` instead of `assertEquals`, right?

build.gradle.kts
deniz-oezdemir: It was included by mistake because we briefly included something like `kotlin.test.assertEquals` from that package but then removed it as JUnit supplies everything we need for testing. Forgot to remove the dependency. Will remove it. :-)

src/test/kotlin/study/DSLTest.kt
deniz-oezdemir: Good catch, I have deleted the study test file and the respective include in the `.gitignore`.

src/main/kotlin/blackjack/model/Dealer.kt
deniz-oezdemir: Thanks for the suggestion!
I’ve extracted the magic numbers into a single `Rules` singleton:

src/main/kotlin/blackjack/model/CardDeck.kt
deniz-oezdemir: I have decided on 4 sets of cards for the deck and changed the initialization logic in `CardDeck` accordingly. :-)

src/main/kotlin/blackjack/model/ResultEvaluator.kt
deniz-oezdemir: Moved all printing logic from `ResultEvaluator` into `OutputView`.
`toString()` now contains only neutral diagnostic info, while `OutputView` has `formatDealer` and `formatPlayer` for console output.

src/main/kotlin/blackjack/model/ResultEvaluator.kt
deniz-oezdemir: Thanks for the source! I'll add a newline at every file end from now on. :-)

src/main/kotlin/blackjack/model/Suit.kt
deniz-oezdemir: Moved all card formatting logic from the model to the `OutputView` layer, so the model remains UI-agnostic.
Now `displayAllCardsMessage` and `displayFirstCardMessage` use helper methods in OutputView (`cardsToString` and `formatCard`) to build the card strings with suit symbols.

src/main/kotlin/blackjack/view/InputView.kt
deniz-oezdemir: Thanks for the suggestion!
I moved the `YES` and `NO` constants to the top of the `InputView` objec.
While it’s true that "y" and "n" are intuitive, I find naming them improves code readability and reduces risk of typos.

src/main/kotlin/blackjack/controller/Controller.kt
deniz-oezdemir: I moved the input retry and error handling logic into the InputView methods.

Also, I moved all player name validation checks into a single `validatePlayerName` function, which is also used when reading user input.

src/main/kotlin/blackjack/controller/Controller.kt
deniz-oezdemir: Thanks for the suggestion! I made the `printResults()` method and a lot of other methods private. I also enabled visibility inspections in IntelliJ to help catch similar cases automatically and keep the code cleaner by restricting method visibility whenever possible.

src/main/kotlin/blackjack/controller/Controller.kt
deniz-oezdemir: You’re right, the extension functions were mixing domain behaviour with UI calls inside the controller. I removed the `Dealer.dealCards()` and `Players.dealCards()` extensions and added methods that replace those to the `Dealer` class.

src/test/kotlin/blackjack/DeckTest.kt
deniz-oezdemir: Yes, I have set my IDE to automatic checking of imports. :-)

src/main/kotlin/blackjack/controller/Controller.kt
deniz-oezdemir: Alright, then I'm leaving it as is for now. :-)

deniz-oezdemir:
Hi @pkeugine, thanks so much for your detailed review! 😄

I have addressed each comment and implemented changes accordingly.

There are still two unresolved change requests/suggestions where I would really appreciate your input before proceeding further. Your perspective would help clarify the best direction there.

I would really appreciate it if you could take another look whenever you have time. Let me know if there is anything else you’d like me to adjust. Thanks again, and I hope you have a great start to the new week! 🙏

deniz-oezdemir:
Hi @pkeugine,
I have just added a Kotlin DSL for constructing Person objects.
I have also added corresponding unit tests to verify each part of the DSL.
This was an additional task following one of the lectures and I would really appreciate if you could review it too.
Thanks, Deniz

src/main/kotlin/blackjack/model/Players.kt
pkeugine: Well, unless there's a special reason behind it (let's say for example, the company or the game system specifically uses that term), I think `Players` is the most intuitive option. :)

src/main/kotlin/blackjack/controller/Controller.kt
pkeugine: Since this is a controller, I agree it might feel harder to write tests.

But if it were part of the domain layer, having private fields wouldn’t necessarily make testing harder than passing variables between functions. In fact, the approach of using fields and injecting them via constructors is called dependency injection. It makes testing much easier by letting you control the dependencies; definitely worth checking out when you have time! It’s actually simpler than it sounds. 🙂

Other than that, great tradeoff analysis, very helpful!

src/main/kotlin/blackjack/view/InputView.kt
pkeugine: Nice, I respect your idea 💯 
Let's keep it that way :)

src/test/kotlin/blackjack/TotalTest.kt
pkeugine: Nice 👍 
I agree, and to top it off, `assertThat()` can chain bunch of neat methods, making it easier to assert things much conveniently.

src/test/kotlin/blackjack/DeckTest.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    @Test
+    fun `test draw more than deck`() {
+        val deck = CardDeck()
+        var card = Card(Rank.ACE, Suit.HEARTS)
+        repeat(208) {
+            card = deck.hit()
+        }
+        assertThrows<Exception> { card = deck.hit() }
+    }
```
pkeugine, line 39: I think we can simplify this test by :  
* remove `var card`. It doesn't look like it's necessary.
* rename it. For example : `throws when drawing more cards than deck contains` since it's asserting for an exception.
* specify the exception : For example : `assertThrows<NoSuchElementException>`. This will allow us to assert the logic with more precision.

Think about this in the next step :)

src/main/kotlin/study/DSL.kt
```diff
@@ -0,0 +1,73 @@
...
+    private val softSkills = mutableListOf<String>()
+    private val hardSkills = mutableListOf<String>()
+
+    fun soft(value: String) {
+        softSkills += value
+    }
+
+    fun hard(value: String) {
+        hardSkills += value
+    }
```
pkeugine, line 70:
```suggestion
    fun soft(vararg values: String) {
        softSkills += values
    }

    fun hard(vararg values: String) {
        hardSkills += values
    }
```

This kind of design would allow us to use it like :
```kotlin
soft("Teamwork", "Speech", "Leadership")
hard("Kotlin", "Gradle")
```

It's just a suggestion though. You did a nice job 👍

src/main/kotlin/study/DSL.kt
```diff
@@ -0,0 +1,73 @@
+package study
```
pkeugine, line 1: It's not a must for now, but there's this thing called [`@DslMarker`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-dsl-marker/) that you can also use. It'll help you make a better DSL. Check it out when you have the time :)

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    fun isBusted(): Boolean = (handCards.total > Rules.BLACKJACK_TARGET)
+
+    fun cardsToString(): String {
+        return handCards.cards.joinToString()
+    }
+
+    fun firstCardToString(): String {
+        return handCards.cards[0].toString()
+    }
```
pkeugine, line 22: Let's remove methods if they aren't used anywhere.

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    private fun formatPlayer(result: PlayerResult): String =
+        when {
+            result.win -> "${result.name}: win"
+            result.draw -> "${result.name}: draw"
+            else -> "${result.name}: loss"
+        }
+
+    // TODO: refactor Pair to value class?
+    fun displayResults(results: Pair<DealerResult, List<PlayerResult>>) {
```
pkeugine, line 58: It'll be nice if we could leave out `TODO`s from PR, and also keep them out of the main branch.
TODO's are mostly for local workspace, and should be implemented if we're trying to merge it to the main branch.

pkeugine (approved):
Hello Deniz,
Nice work going through all the comments 👍 
You've refactored the code very cleanly, and I think we're ready to move on to the next step.

I've added some additional reviews, but they are minor.
Think about them while working on the next step, and we'll talk about it in the next PR if needed 👍 

If you have any questions or opinions, feel free to DM me.

Again, nice work,
and I'll see you soon!
Happy Hacking 🎴 🎴 🎴

src/main/kotlin/blackjack/model/Players.kt
deniz-oezdemir: Cool, then I leave it as is for now. 😄

src/test/kotlin/blackjack/TotalTest.kt
deniz-oezdemir: Thanks for the info. Then I will try to use mainly `assertThat()` in the future.

src/test/kotlin/blackjack/DeckTest.kt
deniz-oezdemir: You are right, I will refactor this test accordingly.

src/main/kotlin/study/DSL.kt
deniz-oezdemir: Thanks for the suggestion! That’s a great idea that makes the DSL more flexible. I have updated the methods accordingly so multiple skills can now be added in a single call. Appreciate this feedback!

src/main/kotlin/blackjack/model/Participant.kt
deniz-oezdemir: Yes, I have realized this also while working on step2 and removed them. Good catch!

src/main/kotlin/blackjack/view/OutputView.kt
deniz-oezdemir: Noted, TODOs are not to be pushed. :-)

src/main/kotlin/study/DSL.kt
deniz-oezdemir: Thanks for the source, it seems like the `DslMarker` makes the DSL safer to use by restricting visibility in nested DSL scopes.

PR #7, opened 2025-07-10
[blackjack step 1] submit Apoorva's mission

Apoorva-S-Natesh:
PR: Step 1 – BlackJack Mission

Hi GuMak,

Thank you for taking the time to review our code.
This pull request includes our implementation for Step 1 of the BlackJack mission. Throughout this mission, we’ve done our best to apply MVC architecture, follow SRP, and practice TDD wherever possible.

We’d be grateful for your feedback on the following areas:

##### 1. MVC Structure
We’ve used Application.kt as the controller. Is this an appropriate use of a controller in the MVC pattern? Additionally, our BlackJack class handles most of the business logic but also calls OutputView. We’re unsure if this compromises the separation of concerns expected in MVC.

##### 2. Code Repetition in Player and Dealer Turns
There is noticeable repetition between the player’s and dealer’s turn logic in the BlackJack class. Is there a more concise or idiomatic Kotlin approach—perhaps using lambda expressions or higher-order functions—that would improve this?

##### 3. Use of Kotlin Language Features
We’ve used features like data class, enum class, and companion objects. Are we using these Kotlin structures appropriately and effectively in this context?

##### 4. Applying DSL (Domain-Specific Language)
In today’s lecture, we were introduced to DSLs in Kotlin. We’re curious how our project could be refactored or extended to incorporate a DSL approach. Any guidance or examples would be very helpful.

##### 5. Abstract Class vs Interface
We created an abstract base class for Player, inherited by both the player and dealer. Would it be more appropriate to use an interface here? We’d appreciate clarification on when to prefer open classes, abstract classes, or interfaces in Kotlin.

We truly appreciate your time and insights. Thank you again for your help in helping us grow.


README.md
```diff
@@ -1 +1,47 @@
...
+- [x] Shuffle cards
+- [x] Pop card from deck
+
+## Hand
+
+- [x] Add card to a hand
+- [x] Calculate scores on hand
+
+## GameController
+
```
Hyeon9mak, line 36: 😉 👍

src/main/kotlin/model/BackJack.kt
```diff
@@ -0,0 +1,55 @@
...
+import view.InputView
+import view.OutputView
+
+class BackJack() {
+    companion object {
+        fun getPlayerNames(): List<String> {
+            val players = InputView.requestPlayerNames()
+            require(players.size <= 6) { "Maximum player names must be 6" }
+            return players
+        }
```
Hyeon9mak: > 3. Use of Kotlin Language Features

If you want to use it in this kind of structure, declaring class as an `object` could be a good option. 🤔

src/main/kotlin/Application.kt
```diff
@@ -0,0 +1,21 @@
...
+    BackJack.runGame(players, dealer)
+    OutputView.displayFinalCardsOnHand(players, dealer)
+    OutputView.displayResults(
+        ResultCalculator.getResult(
+            players.map { it.getScore() },
+            dealer.getScore(),
+        ),
+        players,
+    )
+}
```
Hyeon9mak: > 1. MVC Structure

In the MVC pattern, the Controller acts as a coordinator between the Model and View, and also manages the flow of the Model.
This means that if there are changes to the Blackjack game’s requirements, your teammates should be able to handle all the game rules by simply looking at the Model-related packages.

Currently, `Application` class has minimal knowledge of the Blackjack rules and is mainly focused on communication between the Model and View.
Even though it doesn’t explicitly use the name “Controller,” it seems to be doing its job quite well.

src/main/kotlin/model/BackJack.kt
```diff
@@ -0,0 +1,55 @@
...
+            }
+        }
+
+        private fun dealersTurn(dealer: Dealer) {
+            while (dealer.makeDecision()) {
+                dealer.drawCard(dealer.dealCard())
+                OutputView.displayDealersGame()
+            }
+        }
+    }
```
Hyeon9mak: > 1. MVC Structure

What about the `BlackJack` class?

Whenever the Blackjack rules change, you need to open `BlackJack`.
When the output (View) changes, you also need to open `BlackJack`.
And even when the flow of the Model changes, again, you need to open `BlackJack`.

At the moment, `BlackJack` looks like a giant class that holds the responsibilities of the Model, View, and Controller all at once.
This could lead to maintenance issues down the road.

How about gradually refactoring it—let the `Application` take on the Controller’s role, and hide as much as possible from the View?

src/main/kotlin/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
+package model
+
+class Dealer() : BasePlayer("Dealer") {
+    private val deck = Deck()
+
```
Hyeon9mak: If the name "Dealer" needs to be changed to below image file, which class file will a fellow developer open first: `OutputView`-`Dealer`?

image

src/main/kotlin/model/Suite.kt
```diff
@@ -0,0 +1,8 @@
+package model
+
+enum class Suite(val value: String) {
+    HEARTS("\u2665"),
+    CLUBS("\u2663"),
+    DIAMONDS("\u2662"),
+    SPADES("\u2664"),
+}
```
Hyeon9mak, line 8: When the Blackjack app convert to Web browser image game,

```kotlin
enum class SuitView(
    val suit: Suit,
    val value: String,
) {
    HEART(suit = Suit.HEART, value = "\u2665"),
    // ...
}
```

src/main/kotlin/model/BackJack.kt
```diff
@@ -0,0 +1,55 @@
...
+        ) {
+            players.forEach { player ->
+                repeat(2) {
+                    player.drawCard(dealer.dealCard())
+                }
+            }
+            repeat(2) {
+                dealer.drawCard(dealer.dealCard())
+            }
+        }
```
Hyeon9mak: > 2. Code Repetition in Player and Dealer Turns

It feels like the dealer and player aren’t really collaborating naturally.
Instead, it seems like the `BlackJack` class is just pulling out each of their properties and having them exchange values inside itself.

How about trying to let the dealer and player interact with each other more directly?
If you can combine them well, a lot of the responsibilities currently handled by the `BlackJack` class might naturally disappear.

```kotlin
dealer.selfDrawInitCards()
```

src/main/kotlin/model/Card.kt
```diff
@@ -0,0 +1,22 @@
+package model
+
+// TODO think if we need override all methods
+
+data class Card(val rank: Rank, val suite: Suite, val faceUp: Boolean = true) {
+    override fun equals(other: Any?): Boolean {
+        return rank == (other as Card).rank && suite == other.suite
+    }
```
Hyeon9mak: Was there ever a requirement about faceUp-Down cards? 🤔 
I think this part of the code might confuse other developers when they read it.

Hyeon9mak: > 3. Use of Kotlin Language Features

and the data class already override equals & hashCode.
https://kotlinlang.org/docs/data-classes.html

src/main/kotlin/model/Deck.kt
```diff
@@ -0,0 +1,20 @@
...
+    // for testing
+    fun getCards(): Set<Card> {
+        return cards.toSet()
+    }
+
+    // TODO: Catch the IllegalArgumentException
+    fun pop(): Card {
+        require(cards.isNotEmpty()) { "The deck is empty" }
+        return cards.first().also { cards.remove(it) }
+    }
```
Hyeon9mak: If your testing is done, please remove any code that’s no longer needed.
Also, it’s a good habit to double-check for any leftover TODO comments! 😉

src/main/kotlin/model/Player.kt
```diff
@@ -0,0 +1,14 @@
+package model
+
+import view.InputView
+
+class Player(name: String) : BasePlayer(name) {
+    override fun makeDecision(): Boolean {
+        val decision = InputView.requestPlayerDecision(name)
+        return decision == "y"
+    }
+
```
Hyeon9mak: If the input changes from typing 'y' to clicking a button,
which class do you think your teammates will look at first — the `View` class or the `Player` class?

src/main/kotlin/service/ScoreCalculator.kt
```diff
@@ -0,0 +1,24 @@
...
+                aceCounter--
+            }
+            return scoreTotal
+        }
+
+        private fun hasAce(aceCounter: Int): Boolean = aceCounter != 0
+
+        private fun scoreSum(cards: Set<Card>) = cards.sumOf { card -> Rank.score(card.rank) }
+
+        private fun aceCount(cards: Set<Card>) = cards.count { it.rank == Rank.ACE }
```
Hyeon9mak: It looks like you created a separate class to make testing easier — totally understandable!

I think you could ask Hand to handle this instead, and testing through Hand wouldn’t be too tricky either.
What are your thoughts? 😄

src/test/kotlin/model/DealerTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.Arguments
+import org.junit.jupiter.params.provider.MethodSource
+
+class DealerTest {
+//    @Test
+//    fun `makeDecision returns true if score on hand is 16 or less`() {
+//        val dealer = Dealer()
+//
+//    }
```
Hyeon9mak: 😱

src/main/kotlin/model/BasePlayer.kt
```diff
@@ -0,0 +1,27 @@
+package model
+
+abstract class BasePlayer(val name: String) {
+    protected val hand = Hand()
+
+    init {
```
Hyeon9mak, line 6: > 5. Abstract Class vs Interface

When it’s hard to predict future changes or the direction of your project, using an interface is usually the safer choice.
That’s because, over time, an abstract class you initially designed might no longer align with the future direction of your code, which could eventually force you to refactor it.

However, in cases where there are stable, unchanging rules — like the long-established rules of Blackjack — using an abstract class can be a valid option.

You can easily check whether this approach makes sense by asking simple questions, like through an `Is` check:

- “Is a player a participant in a Blackjack game?”
- “Is a dealer a participant in a Blackjack game?”

Hyeon9mak (changes requested):
Hello Apoorva,
Did a good job to complete Blackjack mission step1 👍

I’ve left replies to your questions along with some code examples.
For each comment, I’ve noted which question it relates to, 
so feel free to take your time going through them.

As for Kotlin DSL, I decided to skip comments on that since I think it’s not a high priority for now.
Kotlin DSL is generally useful and easy to apply in areas like `build.gradle.kts`,
where Groovy-based configuration is common.

But for this mission, I hope you’ll focus mainly on implementing the Blackjack features and practicing the MVC pattern. 😉 

src/main/kotlin/model/BackJack.kt
Apoorva-S-Natesh: Thank you for the feedback! 

I had a quick question regarding this: If a class contains only a single companion object, would it be more appropriate or idiomatic in Kotlin to define it as an object instead? Would the behavior remain the same in that case, or are there subtle differences I should be aware of?

src/main/kotlin/Application.kt
Apoorva-S-Natesh: I now see more clearly how the Controller’s role is to mediate between the View and Model, managing user input and application flow. Given that, it makes sense that the Application class, even without being explicitly named a Controller, is effectively serving that purpose by coordinating between the UI and the game logic. I’ll make sure the Model continues to encapsulate all the game rules so that future changes remain isolated and easy to manage

src/main/kotlin/model/Dealer.kt
Apoorva-S-Natesh: when the design is good, fellow developer should open `OutputView `first not `Dealer`. You are correct and `Dealer.kt` is part of the Model Layer and should not know anything about how its displayed.

**What I did:**
I removed the presentation logic `(toString)` and let the `OutputView` handle it, though dealer is my default named "Dealer" from the constructor at the time of creating instance.

src/main/kotlin/model/Player.kt
Apoorva-S-Natesh: Now I understand what separating concerns means. Thank you for pointing this out. 

**What I did:**
I encapsulated it with an extension function
```
fun String.isYes(): Boolean {
    return this.trim().lowercase() in listOf("y", "yes")
}
```

which will be called by the controller, which now separates logic from view and input from controller.

**Doubt:**
Is it better to delegate interpretation of user input to the Input layer or the extension function?
```
    fun requestPlayerDecision(name: String): Boolean {
        retry("Would $name like to draw another card? (y for yes, n for no)") {
          val input = readln().trim().lowercase()
           return input == "y" || input =="yes"
        }
    }
```

src/main/kotlin/model/BackJack.kt
Apoorva-S-Natesh: This is one of the challenges I faced even during the mission, How to separate concern. Thank you for the tip of how to check if responsibility is mixed id by asking if this feature needs to be changed then what are all the files that should be check and are they all related to the feature? Should they be a place to check even when they are not the responsible file.

**What I did:**
- [x] **BlackJack.kt** had too many responsibilities so I removed logic that was spread across BlackJack and Application into one `Object BlackJackController` which acts as the centralized controller and `Application.kt `now acts as the entry point to the game.

**Doubt:**
The functions inside the controller helping the flow of the game like `startGame()`, `runTurns()` and `showResults()`are private. What’s the best practice for testing such private controller logic? Should I expose them for testing, I read that the if entire flow has to be tested then integration tests are required? Can you please help me what an integration test is where can I learn how to write one?

src/main/kotlin/model/Suite.kt
Apoorva-S-Natesh: you are absolutely right, I had never thought about it this way! 

**What I did:**
Moved all view-specific logic into view package. Model class Card and Suite now only hold domain logic.
I now use:
- `SuiteView` enum to map Suite to unicode (or future image paths)
- `CardView` object to `render` a Card appropriately

src/main/kotlin/model/BackJack.kt
Apoorva-S-Natesh: I Agree! Even though we try to make it close to real world BlackJack the dealer and the Players don't collaborate naturally.

I have refactored to:
- [x] Player Requests Card from Dealer - `player.requestCardFromDealer(dealer)`
- [x] Dealer draws for self - removed repetitive draw logic - dealer.dealFromSelf()

src/main/kotlin/model/Card.kt
Apoorva-S-Natesh: Good point — I understand how that part might seem unclear at first glance. The faceUp/faceDown distinction is currently used to control whether a card should be displayed or hidden, particularly for the dealer’s hidden card during the initial round.

We also considered potential future extensions of the game — for example, if we wanted to support variants where players have one face-down card as well. Having this flag in place makes it easier to adapt the logic without major structural changes. Please suggest me on how to make this intent clearer in the code or through documentation.

src/main/kotlin/model/Deck.kt
Apoorva-S-Natesh: Thank you, Totally missed it but will double check from next time.

**Doubt:**
Are adding comments a good practice? 
Is it followed in real life?
If yes could you please give me an example

src/main/kotlin/model/BasePlayer.kt
Apoorva-S-Natesh: Thank you for the explanation, that gives a good idea if abstract class or Interface should be used. using an abstract class like BasePlayer felt appropriate, especially since both Player and Dealer share common behavior and properties. But if the rules of the game change then using an abstract class might break the code and needs a lot of refactoring, whereas in Interface, for version 1.1 we can add a default function and it offers more flexibility.

src/test/kotlin/model/DealerTest.kt
Apoorva-S-Natesh: Again sorry to have not double checked the code for comments.

src/main/kotlin/service/ScoreCalculator.kt
Apoorva-S-Natesh: Yes, you're absolutely right
The separate class was initially introduced to simplify testing the score logic. But after you mentioned, I’ve refactored the logic into the Hand class itself. This keeps the responsibility where it naturally belongs and still allows for straightforward testing.
Let me know if you think there’s a cleaner way to structure it!

src/main/kotlin/controller/BlackJackController.kt
```diff
@@ -0,0 +1,62 @@
...
+        startGame(players, dealer)
+        runTurns(players, dealer)
+        showResults(players, dealer)
+    }
+
+    private fun getPlayerNames(): List<String> {
+        val players = InputView.requestPlayerNames()
+        require(players.size <= 6) { "Maximum player names must be 6" }
+        return players
+    }
```
Hyeon9mak, line 23: I think the player count limit is also part of the business logic.

What do you think about moving it into a separate class to manage it more clearly?
It could be a good opportunity to introduce an `Players` class as a first-class collection!

src/main/kotlin/controller/BlackJackController.kt
```diff
@@ -0,0 +1,62 @@
...
+
+    private fun startGame(
+        players: List<Player>,
+        dealer: Dealer,
+    ) {
+        dealer.giveInitialCardsToPlayers(players)
+        dealer.drawInitialCards()
+        OutputView.displayInitialCards(players, dealer)
```
Hyeon9mak, line 31: Nice!
lot of the business logic that was exposed in the controller is now nicely hidden. 👍

src/main/kotlin/controller/BlackJackController.kt
```diff
@@ -0,0 +1,62 @@
...
+
+import model.Dealer
+import model.Player
+import service.ResultCalculator
+import util.isYes
+import view.InputView
+import view.OutputView
+
+object BlackJackController {
+    fun playGame() {
```
Hyeon9mak, line 11: > I had a quick question regarding this: If a class contains only a single companion object, would it be more appropriate or idiomatic in Kotlin to define it as an object instead? Would the behavior remain the same in that case, or are there subtle differences I should be aware of?

Since it eventually runs on the JVM, everything gets compiled down to Java anyway
— so there’s no significant difference in terms of functionality.

But because code is also read by people, I think the psychological aspect matters too.
When a teammate sees the keyword object, they can immediately get a sense of the class’s intent.

src/main/kotlin/controller/BlackJackController.kt
```diff
@@ -0,0 +1,62 @@
...
+                player.requestCardFromDealer(dealer)
+                OutputView.displayPlayersTurn(player)
+            }
+        }
+        while (dealer.makeDecision()) {
+            dealer.dealFromSelf()
+            OutputView.displayDealersGame()
+        }
+    }
+
```
Hyeon9mak, line 49: > Can you please help me what an integration test is where can I learn how to write one?

Wanting to test a private method is often a sign that the logic might belong in a separate class.
In this case, I can see that the method is tightly coupled with view logic, so splitting it out won’t be easy — but I believe figuring out how to design objects that are testable and free of view dependencies is part of what makes a strong developer.

Skipping that separation leads to what you mentioned: writing integration tests.
And the most important part of integration testing is asking: What exactly do I want to verify?

Apoorva, what are the key things you’d like to check from the Controller through an integration test?
If you can share those goals first, I’d be happy to think it through with you and help from there.

src/main/kotlin/model/Dealer.kt
```diff
@@ -0,0 +1,25 @@
+package model
+
+class Dealer() : BasePlayer("Deal") {
+    private val deck = Deck()
```
Hyeon9mak, line 4: > though dealer is my default named "Dealer" from the constructor at the time of creating instance.

In a Blackjack game, the dealer's name doesn't seem to be an important part of the business logic.
It's more of a UI-level distinction between the player and the dealer.

I think u can remove `name` property from `BasePlayer` 🤔

src/main/kotlin/view/CardView.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import model.Card
+
+object CardView {
+    fun render(card: Card): String {
+        return if (card.faceUp) {
+            card.rank.value + SuiteView.from(card.suite).symbol
+        } else ""
+    }
+
```
Hyeon9mak, line 11:
```suggestion
object CardView {
    private fun render(card: Card): String {
        return if (card.faceUp) {
            card.rank.value + SuiteView.from(card.suite).symbol
        } else ""
    }

```

😉 👍

src/main/kotlin/model/Dealer.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun giveInitialCardsToPlayers(players: List<Player>) {
+        repeat(2) {
+            players.forEach { player -> player.drawCard(dealCard()) }
+        }
+    }
```
Hyeon9mak, line 10: now It would be easier to write the test code 😉

src/main/kotlin/model/Card.kt
```diff
@@ -0,0 +1,3 @@
+package model
+
+data class Card(val rank: Rank, val suite: Suite, val faceUp: Boolean = true)
```
Hyeon9mak, line 3:
```suggestion
package model

/**
 * Blah Blah Blah
 */
data class Card(val rank: Rank, val suite: Suite, val faceUp: Boolean = true)
```

I think the most reliable, simple, and developer-friendly way is still leaving a comment.
As long as it includes enough context, it’ll help prevent confusion for anyone reading the code later. 👍 

like this
image

src/main/kotlin/model/Deck.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    fun getCards(): Set<Card> {
+        return cards.toSet()
+    }
+
+    // TODO: Catch the IllegalArgumentException
+    fun pop(): Card {
+        require(cards.isNotEmpty()) { "The deck is empty" }
+        return cards.first().also { cards.remove(it) }
+    }
```
Hyeon9mak, line 14: > Are adding comments a good practice?
> Is it followed in real life?

Sure. As you probably know, code alone can’t always express everything
— especially when it comes to business rules or unavoidable team-specific decisions.
In those cases, a good comment can go a long way.

> If yes could you please give me an example

docs comment

image

TODO comment

image

src/main/kotlin/view/InputView.kt
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+    private fun parsePlayerNames(input: String): List<String> {
+        require(input.isNotEmpty() && input.isNotBlank()) {
+            "Input cannot be empty"
+        }
+        return input.split(",").map { it.trim() }
+    }
+
+    private fun parsePlayerDecision(input: String): String {
```
Hyeon9mak, line 21: > Is it better to delegate interpretation of user input to the Input layer or the extension function?

If you imagine a teammate looking for this logic sometime in the future,
which class would they open first?
I think the answer to that question can help guide the right placement. 😄

src/main/kotlin/model/BasePlayer.kt
```diff
@@ -0,0 +1,19 @@
...
+        require(name.isNotEmpty() && name.isNotBlank()) { "Name must not be empty" }
+    }
+
+    fun getScore(): Int = hand.scoreOnHand()
+
+    fun getCardsNumber(): Int = hand.getCards().size
+
```
Hyeon9mak, line 13: 👀 not used!

Hyeon9mak (approved):
Hello Apoorva!
Verrrry great job the reflect the feedback!

I did my best to answer all of your questions in detail.
If you have any follow-up questions, feel free to ask anytime!

I think this step reflects the requirements well,
and the code has been nicely improved — so let’s move on to the step2.
Please take a look at the additional comments as you continue.

Don’t hesitate to reach out if anything’s unclear, and have a great weekend!

src/main/kotlin/controller/BlackJackController.kt
Apoorva-S-Natesh: Thank you making it clear. So when something cannot be tested then there will be an issue with the logic or where and how it is being called.

Here Suppose we were to test if every time the player makes a decision ("y") to draw a card, the size of card on Player's hand increases by 1 (that is the dealer dealt a card to the player)

src/main/kotlin/model/Dealer.kt
Apoorva-S-Natesh: I now see that hardcoding the dealer’s name into the model layer introduces unnecessary coupling between the UI and the core game logic. If the dealer’s label ever needs to change (e.g., localized text or a different role name), that change should ideally be handled at the UI level, not within the model.
Your suggestion to remove the name property from BasePlayer is a good one. It would be cleaner to let only Player instances carry a name, while the Dealer can be distinguished purely through context or presentation logic."

src/main/kotlin/model/Dealer.kt
Apoorva-S-Natesh: Ah! that clarifies my question about integration testing. It makes sense now that the core logic should be tested within the respective classes, while the controller is responsible for coordinating the overall game flow, which is what should be covered by integration tests.

src/main/kotlin/model/Deck.kt
Apoorva-S-Natesh: Thank you for sharing those example comments — they were really helpful

src/main/kotlin/view/InputView.kt
Apoorva-S-Natesh: got it

src/main/kotlin/model/BasePlayer.kt
Apoorva-S-Natesh: Removed the unused function

PR #9, opened 2025-07-10
[Blackjack - Step 1] Submit mehrshahbaz's mission

MehrShahbaz:
- I tried to implement blackjack with the basic knowledge of Kotlin but I would love to get a review for converting the code a medium level kotlin usage,
- Which concepts of Kotlin I could have used to improve the code
- What can I do to make my code be DRY (Do not repeat yourself)
- How can I make this code scalable
- Reduce repetition in my code
- And should I use Inheritance in this Project ?

These are the points I would like to get for my review


README.md
```diff
@@ -1 +1,57 @@
...
+- [] has function `startRound()`
+- [] has function `TakesTurn()`
+  - [] has function `askDrawOrNot()` : until score is greater than 21 or user says 'n'
+
+### class `Statistic`
+- [] will use List<Player> as parameter
+
+
+
+### class `Application`
```
pkeugine: You've written README very well,
but I see some check boxes that aren't checked, and a title with no description.
Let's make sure we mark them all before posting a pull request :)

src/main/kotlin/blackjack/model/GamblerInfo.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model
+
+@JvmInline
+value class GamblerInfo(val name: String) {
+    init {
+        require(name.isNotBlank()) {} // TODO: add err msg
```
pkeugine: It'll be nice to etiher

* implement the TODO before posting a pull request
* or leave the TODO in local instead of merging them into the main branch.

It's really easy to keep something as a TODO, but if it becomes the whole team's habit,
they always stay as a TODO for years or is very difficult to remove them once they join the main branch.
Let's try removing it before this PR gets merged. 👍

src/test/kotlin/blackjack/PlayerTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        val cards =
+            listOf(
+                Card(Rank.TWO, Suit.SPADE),
+                Card(Rank.ACE, Suit.HEART),
+                Card(Rank.KING, Suit.HEART),
+            )
+        val player = Player(GamblerInfo("Player"))
+        player.addCard(cards)
+        assertEquals(13, player.score)
+    }
```
pkeugine: Nice work covering detailed scenarios 👍

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,116 @@
...
+            OutputView.displayCardsOfDealer(dealer)
+            players.forEach { OutputView.displayCardsOfPlayers(it) }
+            OutputView.printEmptyLine()
+            players.forEach { playerTakesTurn(it) }
+            dealerTakesTurn()
+            OutputView.printEmptyLine()
+            OutputView.displayCardsOfPlayersWithScore(dealer)
+            players.forEach { OutputView.displayCardsOfPlayersWithScore(it) }
+            OutputView.printEmptyLine()
+            OutputView.displayFinalResultsHeading()
```
pkeugine: I see consecutive use of `OutputView`.
Why don't we make it so that one `OutputView` method call covers it?
I think `printEmptyLine` can be hidden in individual methods in `OutputView`.

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    fun processPlayerNames(): List<GamblerInfo> {
+        repeat(MAX_ATTEMPTS) {
+            try {
+                val names = InputView.getNamesOfPlayers()
+                return names
+                    .split(",")
+                    .map(String::trim)
+                    .map { it -> GamblerInfo(it) } // to avoid abundant function: caller
```
pkeugine:
```suggestion
                    .map { GamblerInfo(it) }
```

We can remove the arrow in this kind of lambda expression :)
Please check other parts of this project as well.

Also, the above refactored logic can also be:  

```kotlin
                    .map(::GamblerInfo)
```

pkeugine: Just like the `TODO` review,
I don't this comment `to avoid abundant function: caller` is necessary here.
I didn't fully understand what it's trying to say, so it takes more time to understand the
intention. Why don't we remove it, and from now on, try best to let our code explain itself 🙂 .

To go beyond, let's think about the situation where we might actually need comments, and how to write / keep them so that it doesn't get mixed with our logic.

To get some idea, it's great to look into Kotlin and other libraries' implementation.

src/main/kotlin/blackjack/controller/Extensions.kt
```diff
@@ -0,0 +1,9 @@
+package blackjack.controller
+
+fun String.isHitOrStand(): Boolean {
+    return when (this) {
+        "y" -> true
+        "n" -> false
+        else -> throw IllegalArgumentException("The answer must be y or n.")
+    }
+}
```
pkeugine: I'm curious about this extension method.
What are the reasons that you and your pair decided to extend the String class?
I would like you to also tell me the pros and cons that you've felt after this implementation. :)

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,116 @@
...
+            } catch (err: IllegalArgumentException) {
+                OutputView.displayErrorMessages(err.message)
+            }
+        }
+        throw IllegalArgumentException(MAX_ATTEMPT_MESSAGE)
+    }
+
+    private fun dealerTakesTurn() {
+        while (dealer.score <= 16) {
```
pkeugine: Since `BLACKJACK_SCORE` has it's own name, it'll be nice for us to name 16 as well!

pkeugine: It seems like there's a lot of logic in `Controller` that could be divided into different classes,
such as `dealer.score <= 16` and `player.cards.size == 2`.

We can see this by checking the tests. Even though we have a lot of classes in application logic, there's only three test classes covering them. Why don't we try to refactor the `Controller` class so that domain classes hold responsibility of some logic in it?

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,32 @@
...
+    init {
+        generateCards()
+        cards.shuffle()
+    }
+
+    fun drawCard(count: Int = 1): List<Card> {
+        val drawnCards = cards.take(count)
+        removeCards(drawnCards)
+        return drawnCards
+    }
```
pkeugine: This is a small detail, but since we're returning a List for this method, naming it `drawCards` will help the developer who uses this method understand how it works more clearly.

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    init {
+        generateCards()
+        cards.shuffle()
+    }
+
+    fun drawCard(count: Int = 1): List<Card> {
+        val drawnCards = cards.take(count)
+        removeCards(drawnCards)
+        return drawnCards
```
pkeugine: What if there aren't enough cards to draw? 🤔

src/main/kotlin/blackjack/model/Suit.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model
+
+enum class Suit(val symbol: String) {
+    SPADE("♠"),
+    HEART("♥"),
+    CLUB("♣"),
+    DIAMOND("♦"),
```
But here's my thought:

Since the visual effects are used in console based output, I think they belong to the `OutputView`.

src/main/kotlin/blackjack/view/InputPrompt.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.view
+
+object InputPrompt {
+    const val GET_NAME_OF_PLAYER = "Enter the names of the players (comma-separated):"
+
+    fun askHitOrStand(name: String): String = "Would $name like to draw another card? (y for yes, n for no)"
+}
```
pkeugine: What are the pros and cons of separating `InputPrompt` from `InputView`?
If you find it having more pros than cons, I expect an `OutputPrompt` as well.
How does this sound?
(This is more of a questions rather than suggestion!)

src/test/kotlin/blackjack/GamblerInfoTest.kt
```diff
@@ -0,0 +1,27 @@
...
+            GamblerInfo("")
+        }
+    }
+
+    @Test
+    fun `should throw exception - blank input`() {
+        assertThrows<IllegalArgumentException> {
+            GamblerInfo("    ")
+        }
+    }
```
pkeugine: I think we can utilize `ParameterizedTest` for these tests.
This way, we can easily manage invalid name inputs when there's more.
Why don't we give it a try!

README.md
```diff
@@ -1 +1,57 @@
-# kotlin-blackjack
\ No newline at end of file
+# Kotlin Blackjack
```
pkeugine, line 1: **Q. should I use Inheritance in this Project ?**

I would like to know which part of this project you thought of inheritence, and why.
Even better, your attempt on it with examples!
Then, we'll be able to communicate in much detail 👍

README.md
```diff
@@ -1 +1,57 @@
...
+
```
pkeugine, line 2: **Q. Reduce repetition in my code**
**Q. What can I do to make my code be DRY (Do not repeat yourself)**

I believe these two questions are similar to each other, and I'd like to ask you first:
`Why is DRY important? Is is always necessary?`

It'll be fun to talk about this with other crews or pair. :)

README.md
```diff
@@ -1 +1,57 @@
...
+## Features
+
```
pkeugine, line 4: **Q. How can I make this code scalable**

Kind of a vague question I must say. 🤔 

Fun way to think about it is to ask yourself:
In which way do you think this project could get expanded?
More rules? Multiplayer? Any other output method other than console?

Tell me about your thought :)

pkeugine (changes requested):
Hello Shahbaz, and welcome to mission number 2. 🙌 
I see you thought of detailed cases in card count and added tests for them 👍 

I added some comments for you to check out,
so once you're done with them, hit me with another review request.

Also, if you have any more questions or thoughts, leave them in this PR!
Or you could DM me if you think the questions are personal 😉 

Keep up the good work, and
Happy hacking! 💻 💻 💻

src/main/kotlin/blackjack/controller/Extensions.kt
MehrShahbaz: It was more like a whim, that we need a method to manipulate string but now looking back I think this could have been not a good idea as it is a helper function for the app not the String class I think I should move it to a method in controller

README.md
MehrShahbaz: The Idea which I going with does not require inheritance as I consider gambler and dealer as a player, but just for my understanding I asked this question, but now thinking I feel like the type of Modals I created does not require Inheritance
But I would like to try that for learning purpose

README.md
MehrShahbaz: To extract the logic to a different class or wrapper can make it scalable as new rules can be added to final result to calculate the final results and trying to make each chunk of code independent can help achieve that

src/main/kotlin/blackjack/view/InputPrompt.kt
MehrShahbaz: It can help us to change all the View things in one place more like breaking the string and calculation logic, in a bit project it can help us to go for translation as if we change the language without going through the output logic

README.md
pkeugine: If you don't think it's necessary but want to try it out for learning purpose,
you can make another branch off of this current one and try it out there.
As for my opinion, I do agree with you that it's not a 'must' in this project.
I would use a different approach such as making a `Role` enum which I personally think is an easier method and is easier to maintain.

README.md
pkeugine: Nice! It's not always the right way to achieve scalable, but it is highly likely.
I guess you know how to make a scalable project already! 😄

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,123 @@
...
+        return when (input) {
+            "y" -> true
+            "n" -> false
+            else -> throw IllegalArgumentException("The answer must be y or n.")
+        }
+    }
+
+    companion object {
+        private const val MAX_ATTEMPTS = 5
+        const val BLACKJACK_SCORE = 21
```
pkeugine, line 119: If this magic number is only used in `FinalResult`,
we can move this to `FinalResult` and maintain it there.

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,123 @@
...
+        var answer = false
+        while (player.isPlayerBelowBlackJack()) {
+            answer = processHitOrStay(player)
+            if (!answer) {
+                break
+            }
+            player.addCard(deck.drawCards())
+            OutputView.displayCardsOfPlayers(player)
+        }
+        if (!answer && player.cards.size == INITIAL_CARD_COUNT) {
```
pkeugine, line 62:
```suggestion
        if (!answer && player.hasCardCount(INITIAL_CARD_COUNT)) {
```

In the previous comment, this was my intended suggestion!

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    private fun processPlayerNames(): List<GamblerInfo> {
+        repeat(MAX_ATTEMPTS) {
+            try {
+                val names = InputView.getNamesOfPlayers()
+                return names
+                    .parseCommaString()
+                    .map(::GamblerInfo)
```
pkeugine, line 90: Nice job refactoring the logic.
Excellent use of an extension function as well. 👍

src/main/kotlin/blackjack/model/FinalResult.kt
```diff
@@ -0,0 +1,24 @@
+package blackjack.model
+
+import blackjack.controller.Controller.Companion.BLACKJACK_SCORE
+
+class FinalResult(val dealer: Dealer, players: List<Player>) {
```
pkeugine, line 5:
```suggestion
class FinalResult(dealer: Dealer, players: List<Player>) {
```

I think we can remove the `val` here :)

src/main/kotlin/blackjack/model/FinalResult.kt
```diff
@@ -0,0 +1,24 @@
...
+    var win: List<Player> = listOf()
+        private set
+    var lose: List<Player> = listOf()
+        private set
+    var draw: List<Player> = listOf()
+        private set
```
pkeugine, line 11:
```suggestion
    val win: List<Player>
    val lose: List<Player>
    val draw: List<Player>
```

What do you think are the pros and cons of these two implementations?

src/test/kotlin/blackjack/GamblerTest.kt
```diff
@@ -0,0 +1,23 @@
+package blackjack
+
+import blackjack.model.*
```
pkeugine, line 3: I recommend avoiding `wildcard`(`*`) in imports.
Can you guess the reason behind it? :)

pkeugine (approved):
Hello Shahbaz,
and that was quick!
Not only quick, you've refactored your project very well.

I've added some simple additional reviews.
Since they seem quite minor, let's work on it as we move on to the next step.

If you have any questions, feel free to add a comment here or tell me about it in the next PR message. Or you could send me a DM.

As always,
Happy Hacking! 🛵 🛵 🛵

PR #10, opened 2025-07-10
[blackjack - Step1] Submit AnnLvu's mission 

AnnLvu:
# Step 1 PR

Hi!  
That's my step1 PR.  
I’d like to thank you in advance for your time spent on the review and for any future advice.

This was a really interesting mission for me. I tried using many new things: an interface, parameterized tests, and lambda expressions. And I played blackjack for the first time! :) I learned so much!  
I tried to apply the MVC pattern based on my current understanding, with a strong focus on separating the responsibilities of the main controller. That led me to create multiple manager classes to handle different parts of the logic. I also aimed to follow proper class design principles by limiting access between classes and objects.  
Although TDD is still challenging to fully integrate, I'm starting to get used to the mindset. Overall, I’m happy to see clear improvement compared to my previous project in many areas.

We tried to optimize everything as much as possible, but there were still a few things we couldn't simplify. We left TODOs in two places. Maybe you could give me some advice on how I could improve them:  
- **Controller -> PlayerManager -> `fun askPlayerHit`**  
- **utils -> CardGenerator -> `fun generateSuits()`**

---

## Controller

- **GameManager:** Manages the overall game flow. Responsibilities will be split into manager classes  
- **CardManager:** Handles deck management using `ArrayDeque`, tracks used/available cards  
- **PlayerManager / DealerManager / StatsManager:** Extracted from GameManager to improve separation of concerns

## model
- **Card:** Represents a playing card with a name and a digit  
- **Hand:** Stores a list of cards and provides a method to calculate the hand’s score  
- **Rank:** Enum used for scoring hands based on card ranks  
- **Playable Interface:** Introduced for shared logic between Player and Dealer  
- **Player / Dealer:** Implementations of Playable. Dealer includes logic to decide whether to draw based on rules
-  **Stats:** Tracks individual player results and overall dealer performance 

## view

- **InputView:** Handles user inputs like player names and draw decisions. Includes validation  
- **OutputView:** Displays game state — initial hands, card draws, final results, and dealer actions


README.md
```diff
@@ -1 +1,115 @@
...
+- [x] has method `requestCard()` to request a card by condition then, ask the card manager with boolean
+- [x] has method `drawCard()` to take card given by a card manager
+- [x] has method `calculateHand()` to calculate score of hand
+- [x] has method `isBust()` to figure out the player of the dealer is bust or not
+
+Player: Playable
+
+- [x] follows interface `Playable`
+- [x] has method `calculateHand()` to calculate score of hand with `Rank` of `Card`
+    - TODO: MUST have to find a solid logic for calculation
```
Hyeon9mak, line 33: 😉 👍

src/main/kotlin/blackjack/controller/CardManager.kt
```diff
@@ -0,0 +1,46 @@
...
+        _cardMap = map.toMap()
+    }
+
+    private fun getCard(): Card {
+        val deque = ArrayDeque(cards)
+        val card = deque.first()
+        deque.removeFirst()
+        _cards = deque.toList()
+        return card
+    }
```
Hyeon9mak: I think the Controller's role in the MVC pattern is to help communication between Model(Domain) and View.
That means all business rules should be baked into the Model(Domain), so when business policies change, fellow developers only need to look at the Model(Domain) package.

Since the Controller helps communication between Model and View,
 I think it can naturally manage the business flow to some extent. 

But the current structure has the Controller pulling out all properties from the Model and handling everything one by one.
I'm guessing writing test code was pretty challenging, wasn't it? 🤔

Hyeon9mak: How about trying to let the Model classes interact with each other more directly?
Then naturally, the need for Manager classes will gradually disappear. 😉 

```kotlin
dealer.selfDrawInitCards()
```

src/main/kotlin/blackjack/controller/PlayerManager.kt
```diff
@@ -0,0 +1,48 @@
...
+        receiveCard: () -> Card,
+    ) {
+        var isFirst = true
+        while (!player.isBust()) {
+            val requestMessage =
+                player.requestCard {
+                    callGameManagerToReadYesOrNo(player.name)
+                }
+            if (requestMessage) {
+                isFirst = false
```
Hyeon9mak: How about make negative asking method?
It can be remove `!` operator. `!` operator make lot of mistake.

src/main/kotlin/blackjack/model/Card.kt
```diff
@@ -0,0 +1,6 @@
+package blackjack.model
+
+data class Card(val name: String) {
+    val digit = name.substring(0, name.length - 1)
+    val value = Rank.of(digit).value
+}
```
Hyeon9mak: I guess u want like this...

```kotlin
data class Card(name: String) {
    // ...
}
```

but, Does Card really need to know about strings? 🤔

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,22 @@
+package blackjack.model
+
+data class Dealer(override val name: String = "Dealer") : Playable {
+    private var _hand = Hand()
+    override val hand: Hand
+        get() = _hand
```
Hyeon9mak: If the name "Dealer" needs to be changed to below image file, which class file will a fellow developer open first: `OutputView`-`Dealer`?

image

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,22 @@
+package blackjack.model
+
+data class Hand(val cards: List<Card> = emptyList()) {
```
Hyeon9mak, line 5: What's different between data class & value class?
- https://kotlinlang.org/docs/inline-classes.html
- https://kotlinlang.org/docs/data-classes.html

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,22 @@
...
+        var aceCounts = values.count { it == 11 }
+        var score = values.sum()
+        while (aceCounts > 0 && score > 21) {
+            score -= 10
+            aceCounts--
+        }
+        if (score < 10 && values.contains(11)) {
+            score += 10
+        }
+        return score
```
Hyeon9mak: I think u can give a new name for 21 score. 😄

src/main/kotlin/blackjack/model/Playable.kt
```diff
@@ -0,0 +1,18 @@
+package blackjack.model
+
+interface Playable {
+    val name: String
+    val hand: Hand
```
Hyeon9mak: What's difference between interface & abstract class?

src/main/kotlin/blackjack/model/Rank.kt
```diff
@@ -0,0 +1,25 @@
...
+    SEVEN("7", 7),
+    EIGHT("8", 8),
+    NINE("9", 9),
+    TEN("10", 10),
+    JACK("J", 10),
+    QUEEN("Q", 10),
+    KING("K", 10),
+    NONE("0", 0),
+    ;
+
```
Hyeon9mak: When the Blackjack app convert to Web browser image game,

```kotlin
enum class SuitView(
    val suit: Suit,
    val symbol: String,
) {
    // ...
}
```

src/main/kotlin/blackjack/model/Stats.kt
```diff
@@ -0,0 +1,48 @@
+package blackjack.model
+
+data class Stats(val players: List<Player>, val dealer: Dealer) {
+    private var _playerBoard = initPlayerBoard()
+    val playerBoard: Map<Player, Int> get() = _playerBoard
+
+    private var _dealerStats = mapOf("win" to 0, "lose" to 0, "tie" to 0)
+    val dealerStats: Map<String, Int> get() = _dealerStats
+
```
Hyeon9mak: I think those are View's responsibility. 😮 
How do u think?

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    fun readYesOrNo(name: String): Boolean {
+        println("\nWould $name like to draw another card? (y for yes, n for no)")
+        val input = readln().trim()
+        return when (input) {
+            "y" -> true
+            "n" -> false
+            else -> throw IllegalArgumentException("Please write 'y' for 'yes' or 'n' for 'no'.")
+        }
+    }
```
Hyeon9mak: Yeah! `y`, `n` strings are not important things of Blackjack game.

but.. 😱 

image

Hyeon9mak (changes requested):
Hello Ann!
Nice work to complete Blackjack mission step1 👍

I was really impressed by your attempt to split the Controller into multiple parts to distribute responsibilities. 
Since HeroTechCourse is all about trying, experiencing, and improving different ways to write maintainable code, I want to encourage these kinds of attempts.

Unfortunately, through this experience, you probably noticed that while Controller responsibilities were distributed, the Model (Domain) classes became too lightweight as a downside. 
To solve this problem, I think we need to change the approach from pulling properties out of Model (Domain) classes to handle them in Controllers, to having Model (Domain) classes communicate directly with each other.

I've left some related comments, so please take your time checking them out and feel free to ask questions if you run into any difficulties.

Feel free to ask any questions u have! 😉

src/main/kotlin/blackjack/model/Playable.kt
AnnLvu: Yeah, originally Playable was an interface because it felt like `Player` and `Dealer` were just different participants in the game, each with potentially unique behavior.
Player and Dealer actually share a lot of the same logic .So it's not just similar structure - they literally do the same things in many places.
That’s why I changed `Playable` from an interface to an abstract class. It allowed me to centralize shared behavior, reduce duplication, and make the architecture cleaner and easier to maintain.
An **interface** is more about what something can do, but doesn't say how. Interfaces are great when you only need to define a contract.
An **abstract class** lets you define both what and how, which fits our case better.

src/main/kotlin/blackjack/view/InputView.kt
AnnLvu: I’ve changed it -now, every time the user enters an incorrect "y" or "n", there’s an opportunity to enter the answer again right away, and only then does the next message appear.

src/main/kotlin/blackjack/model/Stats.kt
AnnLvu: I moved it to the View folder and renamed it to StatsView!

src/main/kotlin/blackjack/model/Hand.kt
AnnLvu: Create a new object Constants. Thank you!

src/main/kotlin/blackjack/model/Rank.kt
AnnLvu: Thanks for the suggestions! I've added `SuitView` and `RankView` in `blackjack.view` for display logic

src/main/kotlin/blackjack/model/Hand.kt
AnnLvu: Thanks for the links and question!
I kept `Hand` as a `data class` because it holds a `List<Card>` and relies on auto-generated `toString()` for terminal output in `OutputView`. The `calculateCards()` method also adds behavior, making `Hand` a complex entity.

src/main/kotlin/blackjack/model/Dealer.kt
AnnLvu: Previously, to change the dealer's name, developer had to modify OutputView manually.
Now, the dealer’s display name is handled directly inside the Dealer class, so developers only need to update Dealer - no need to touch OutputView anymore.

src/main/kotlin/blackjack/controller/PlayerManager.kt
AnnLvu: Create new method fun canContinue(): Boolean

src/main/kotlin/blackjack/model/Card.kt
AnnLvu: You're right to ask! I kept Card(rank, suit) to avoid string parsing logic inside the model

src/main/kotlin/blackjack/controller/CardManager.kt
AnnLvu: Changed it! Thank you!

src/main/kotlin/blackjack/model/Playable.kt
```diff
@@ -0,0 +1,32 @@
...
+        return handInternal.calculateCards()
+    }
+
+    fun isBust(): Boolean {
+        return calculateHand() > Constants.BLACK_JACK
+    }
+
+    fun canContinue(): Boolean {
+        return calculateHand() <= Constants.BLACK_JACK
+    }
```
Hyeon9mak, line 31: 😉 👍

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,27 @@
+package blackjack.model
+
+import blackjack.controller.Deck
+
+data class Dealer(override val name: String = "Dealer") : Playable() {
+    fun shouldDrawCardOrNot(): Boolean {
+        val score = calculateHand()
+        return score <= 16
+    }
+
```
Hyeon9mak, line 10: What do you think is the role of the View in the MVC pattern?
Could you explain how you see the responsibilities being divided between the Model and the View?
I think that would help me better understand your intent. 💭

src/main/kotlin/blackjack/model/Playable.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    open fun requestCard(condition: () -> Boolean): Boolean {
+        return condition()
+    }
+
+    open fun drawCard(newCard: Card) {
+        val deque = ArrayDeque(hand.cards)
+        deque.addLast(newCard)
+        handInternal = Hand(deque.toList())
+    }
```
Hyeon9mak, line 19: To put it more intuitively, you can decide between using an interface or an abstract class based on whether the rules of the real-world concept are likely to change.

For example, the rules of Blackjack have been passed down for over a hundred years and are unlikely to change anytime soon. Based on the assumption that "both the dealer and the player are participants," it's reasonable to use an abstract class.

On the other hand, if a component is difficult to predict in terms of how it might change or expand, it's better to use an interface to leave room for flexibility and future extensions.

src/main/kotlin/blackjack/view/SuitView.kt
```diff
@@ -0,0 +1,23 @@
...
+        }
+
+    fun fromSymbol(symbol: Char): Suit =
+        when (symbol) {
+            '♥' -> Suit.HEARTS
+            '♦' -> Suit.DIAMONDS
+            '♣' -> Suit.CLUBS
+            '♠' -> Suit.SPADES
+            else -> Suit.NONE
+        }
```
Hyeon9mak, line 22: I think it's only used in test code 🤔

src/main/kotlin/blackjack/controller/PlayerManager.kt
```diff
@@ -0,0 +1,48 @@
...
+        val mutableList = _players.toMutableList()
+        val player = Player(name)
+        mutableList.add(player)
+        _players = mutableList.toList()
+    }
+
+    // TODO: resolve indent depth
+    fun askPlayerHit(
+        player: Player,
+        receiveCard: () -> Card,
```
Hyeon9mak, line 20: need double check the TODO comments!

src/main/kotlin/blackjack/Application.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack
+
+import blackjack.controller.GameManager
+
+fun main() {
+    val gm = GameManager
+    gm.run()
+}
```
Hyeon9mak, line 8: image

src/main/kotlin/blackjack/controller/DealerManager.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.controller
+
+import blackjack.model.Dealer
+
+class DealerManager {
+    val dealer = Dealer()
+}
```
Hyeon9mak, line 7: is this really need class? 🤔

src/main/kotlin/blackjack/model/Hand.kt
Hyeon9mak: Doesn't it the `value class` also auto-generate `toString()`? 🤔

Hyeon9mak (approved):
Hi Ann,
Great work incorporating the feedback from step1!

There are a few comments where I left follow-up questions because I wasn’t fully sure about your intentions. 
When you submit the PR for step2, it would be great if you could also include your answers to those questions.
— that way, I can give you more specific feedback.

And of course, if you have any questions yourself, feel free to ask anytime!

Enjoy the rest of your weekend! 😊

src/main/kotlin/blackjack/controller/DealerManager.kt
AnnLvu: I suppose, no. I delete DealerManager and use Dealer directly in GameManager

src/main/kotlin/blackjack/Application.kt
AnnLvu: Rename dm -> gameManager

src/main/kotlin/blackjack/controller/PlayerManager.kt
AnnLvu: OMG, I forgot about this, sorry!

src/main/kotlin/blackjack/model/Playable.kt
AnnLvu: Thank you so much for explanation!

src/main/kotlin/blackjack/model/Hand.kt
AnnLvu: Hi! Thanks for the comment and question about value class and toString()!  I decided to keep Hand as a data class and retain the toText() method for the following reasons:
- Hand is not just a wrapper around List<Card> but an entity with additional logic (calculateCards() for scoring with ace handling). A data class is better suited for such cases, as it supports complex data structures and auto-generates useful methods
- The toText() method provides a clean, user-friendly output (A♥, K♠) that’s perfect for the game’s console interface. The auto-generated toString() from a data class outputs Hand(cards=[A♥, K♠]) - in our Example Output we have (A♥, K♠)

PR #11, opened 2025-07-10
[Blackjack - step1] Submit saragrosser's mission 

saragrosser:
This PR completes the first step of the Kotlin Blackjack project: a terminal-based simplified Blackjack game where players compete against a dealer. The game includes full dealing logic, draw rules, score calculation (with Ace logic), and win/loss/draw result handling.

**What's Included:**

-Core classes for Card, Deck, Participant, Player, and Dealer

-Game controller logic (Game.kt) to run a full game flow

-Score handling extracted into ScoreCalculator.kt

-Result logic extracted into GameLogic.kt and returned via a GameResult enum

-Console input/output using InputView and OutputView

-Unit tests 

All your feedback is welcome thank you so much!


README.md
```diff
@@ -1 +1,28 @@
...
+- [x] Calculate participant score with Ace logic
+- [x] Announce game results
+
+### Game
+- [x] Accept player names
+- [x] Deal two initial cards to all participants
+- [x] Handle player turns
+- [x] Handle dealer turn
+- [x] Display final hands and scores
+- [x] Determine and print winners using `GameLogic.getGameResult()`
```
Hyeon9mak, line 28: 😉 👍

src/main/kotlin/blackjack/Application.kt
```diff
@@ -0,0 +1,5 @@
+package blackjack
+
+fun main() {
+    Game().run()
+}
```
Hyeon9mak: https://medium.com/@hyeon9mak/about-no-newline-at-end-of-file-warning-on-github-b42d14dccc93 😉

src/main/kotlin/blackjack/Card.kt
```diff
@@ -0,0 +1,38 @@
...
+    FIVE("5", 5),
+    SIX("6", 6),
+    SEVEN("7", 7),
+    EIGHT("8", 8),
+    NINE("9", 9),
+    TEN("10", 10),
+    JACK("J", 10),
+    QUEEN("Q", 10),
+    KING("K", 10),
+    ACE("A", 11)
```
Hyeon9mak: When the Blackjack app convert to Web browser image game,

```kotlin
enum class SuitView(
    val suit: Suit,
    val symbol: String,
) {
    // ...
}
```

src/main/kotlin/blackjack/Card.kt
```diff
@@ -0,0 +1,38 @@
...
+        return "${rank.symbol}${suit.symbol}"
+    }
+
+    fun getScore(): Int {
+        return rank.value
+    }
+
+    fun isAce(): Boolean {
+        return rank == Rank.ACE
+    }
```
Hyeon9mak: > Do not use getter, Ask to Object.

Nice work! 👍 
When the `Rank` class chagned, outer classes does not be affected!

src/main/kotlin/blackjack/Dealer.kt
```diff
@@ -0,0 +1,32 @@
+package blackjack
+
+class Dealer(private val deck: Deck) : Participant("Dealer") {
```
Hyeon9mak: If the name "Dealer" needs to be changed to below image file, which class file will a fellow developer open first: `OutputView`-`Dealer`?

image

src/main/kotlin/blackjack/Dealer.kt
```diff
@@ -0,0 +1,32 @@
...
+            val wantsToDraw = InputView.askPlayerWantsToDraw(player.name)
+            if (wantsToDraw) {
+                giveCardTo(player)
+                OutputView.displayPlayerHand(player)
+                if (ScoreCalculator.calculate(player) >=21) break
+            } else {
+                break
+            }
+        }
+    }
```
Hyeon9mak: You’ve probably learned about the MVC(Model-View-Controller) pattern in a Lotto mission, right? 😉 

I think All HeroTechCourse missions are designed under the assumption that
'the domain doesn’t change often, but the view changes frequently.'.

Your current code, the domain and view depend on each other.
When that happens, changes in one part will keep affecting the other.

How could you improve your code so that the domain isn’t affected by changes in the view? 💭

src/main/kotlin/blackjack/Deck.kt
```diff
@@ -0,0 +1,21 @@
...
+            for (rank in Rank.values()) {
+                cards.add(Card(suit, rank))
+            }
+        }
+        cards.shuffle()
+    }
+    fun drawCard(): Card {
+        if (cards.isEmpty()) throw IllegalStateException("Deck is empty")
+        return cards.removeFirst()
+    }
```
Hyeon9mak: Niiiiiiiiiice 💯 👍

src/main/kotlin/blackjack/Game.kt
```diff
@@ -0,0 +1,60 @@
...
+
+class Game {
+
+    fun run() {
+        val playerNames = InputView.askPlayerNames()
+        val players = playerNames.map { Player(it) }
+        val deck = Deck()
+        val dealer = Dealer(deck)
+
+        OutputView.displayDealing(playerNames)
```
Hyeon9mak: Also possible for the `Dealer` create and hold the `Deck` inside constructor.
It's more easier maintain the code.

src/main/kotlin/blackjack/Game.kt
```diff
@@ -0,0 +1,60 @@
...
+
+        repeat(2) {
+            dealer.giveCardTo(dealer)
+            players.forEach { player ->
+                dealer.giveCardTo(player)
+            }
+        }
```
Hyeon9mak: How about this? 😉 
I think it can make more easier to generate test code.

```kotlin
dealer.giveTwoCardsTo(players)

"dealer give two cards to all players" {
    // ...
}
```

src/main/kotlin/blackjack/Game.kt
```diff
@@ -0,0 +1,60 @@
...
+        OutputView.displayDealerFirstTwoCards(dealer.getHand().take(2))
+        players.forEach { OutputView.displayPlayerHand(it) }
+
+        println()
+
+        players.forEach { dealer.askPlayerDraw(it) }
+
+        println()
+        players.forEach { OutputView.displayPlayerHand(it) }
```
Hyeon9mak: > When the Blackjack app convert to Web browser image game

is it really the `Game`'s responsibility to call `println()`? 🤔

src/main/kotlin/blackjack/InputView.kt
```diff
@@ -0,0 +1,20 @@
...
+        val input = readln()
+        return input.split(",").map { it.trim() }.filter { it.isNotEmpty() }
+    }
+    fun askPlayerWantsToDraw(playerName: String): Boolean {
+        println("Would $playerName like to draw another card? (y for yes, n for no)")
+        return when (readln().trim().lowercase()) {
+            "y" -> true
+            "n" -> false
+            else -> {
+                println("Invalid input. Please enter 'y' or 'n'.")
```
Hyeon9mak: Niiiice. `y`, `n` strings are not important things of Blackjack game.

src/main/kotlin/blackjack/ScoreCalculator.kt
```diff
@@ -0,0 +1,21 @@
...
+        val hand = participant.getHand()
+        var total = hand.sumOf { it.getScore() }
+        var aces = hand.count { it.isAce() }
+
+        while (total > 21 && aces > 0) {
+            total -= 10
+            aces--
+        }
+
+        return total
```
Hyeon9mak: How about ask "total score" and "it had Ace card" to `Hand`?

> Do not use getter.

src/test/kotlin/blackjack/DeckTest.kt
```diff
@@ -0,0 +1,34 @@
...
+        assertEquals(51, deck.remainingCards())
+    }
+
+    @Test
+    fun `should throw error when drawing from empty deck`() {
+        repeat(52) { deck.drawCard() }
+        assertThrows(IllegalStateException::class.java) {
+            deck.drawCard()
+        }
+    }
```
Hyeon9mak: Very detail test cases. nice work! 💯

Hyeon9mak (changes requested):
Hello Sari, 
Did a great job to complete Blackjack mission step1 👍 

I focused most of my comments on the MVC pattern.
It probably won’t be too difficult for you to work through them. 😄 

src/test/kotlin/blackjack/DeckTest.kt
saragrosser: Thank you!!

src/main/kotlin/blackjack/Application.kt
saragrosser: Oops! thank you!! I will be more aware of the styling conventions!

src/main/kotlin/blackjack/Card.kt
saragrosser: Hi!! thank you for your feedback, I have moved view symbols to SuitView and RankView to have all view logic inside the view package

src/main/kotlin/blackjack/Card.kt
saragrosser: Thank you!!

src/main/kotlin/blackjack/InputView.kt
saragrosser: I totally agree—those input values are just a UI detail and not core to the game logic. Appreciate the confirmation! 😊

src/main/kotlin/blackjack/Game.kt
saragrosser: Hi, you are right it is not Game responsibility, it should be in view logic so I have moved  all "println" and display logic from Game to OutputView :)

src/main/kotlin/blackjack/Game.kt
saragrosser: Yes, it makes sense that the Dealer is the one who creates the Deck, which also makes it unnecessary to create a Deck every time separately. Thank you!

src/main/kotlin/blackjack/Game.kt
saragrosser: You are totally right! I moved this logic to the Dealer, and now the code inside Game is more readable. Thank you!

src/main/kotlin/blackjack/ScoreCalculator.kt
saragrosser: Yes, this makes total sense, and after extracting the logic to a Hand class, I needed to refactor the code even more which resulted in a more readable code! Thank you!

src/main/kotlin/blackjack/Dealer.kt
saragrosser: Regarding your question about which file a developer would open if we needed to change how the "Dealer" is represented (like in the image), they would now open the `OutputView.kt` file, not `Dealer.kt`.

All display logic (such as showing the dealer’s cards or describing their actions) has been moved to the view layer. The `Dealer.kt` class only contains domain logic, so it's unaffected by presentation changes. This separation follows MVC principles and keeps the domain isolated from external changes.

Thanks again for helping me see this separation more clearly! 🙂

src/main/kotlin/blackjack/Dealer.kt
saragrosser: You're absolutely right, the original `Dealer.askPlayerDraw()` method was tightly coupling the domain logic with the view.

I’ve now decoupled all view-related responsibilities by moving those interactions to `OutputView.kt`. The domain layer is now only concerned with game logic, and doesn't directly reference any input/output. This ensures the logic can evolve independently of the view, following MVC principles more closely.

To reflect this separation more clearly, I also refactored the folder structure into:
- `model` → all domain entities (Dealer, Player, Card, etc.)
- `view` → all terminal display logic (InputView, OutputView, SuitView, etc.)
- `controller` → orchestrating logic (GameLogic, GameController)

That said, I had a follow-up question:  
From what I’ve read, when not building an app with clear API endpoints (like a web controller), some recommend using a `service` layer instead of a `controller` directory to hold the logic (e.g., `GameController`). Would that be more appropriate in this case?

Thanks again for your guidance, this step really helped me understand how to maintain a clean architecture and makes it more clear to me for future project implementations.

정말 감사합니다!

src/main/kotlin/blackjack/view/SuitView.kt
```diff
@@ -0,0 +1,16 @@
+package blackjack.view
+
+import blackjack.model.Suit
+
+enum class SuitView(val suit: Suit, val symbol: String) {
+    HEART(Suit.HEART, "♥"),
+    DIAMOND(Suit.DIAMOND, "♦"),
+    CLUB(Suit.CLUB, "♣"),
```
Hyeon9mak, line 8: I think u can change access modifier of `Suit` to private.
Those only use in companion object area.

src/main/kotlin/blackjack/model/GameResult.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.model
+
+enum class GameResult(val label: String) {
+    WIN("Win"),
+    LOSE("Lose"),
+    DRAW("Draw")
+}
```
Hyeon9mak, line 7: And these labels are looks like Rank and Suit 😄

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,46 @@
...
+import blackjack.view.OutputView
+
+class Dealer : Participant("Dealer") {
+
+    private val deck = Deck()
+
+    fun giveTwoCardsTo(participants: List<Participant>) {
+        repeat(2) {
+            giveCardTo(this)
+            participants.forEach { giveCardTo(it) }
```
Hyeon9mak, line 13: model class still know view package classes. 😱 

How about reversing the direction
— instead of the model using the view, let the view use the model instead?
The controller can help coordinate that interaction in between!

Hyeon9mak: And Let's think about Dealer's name.
In blackjack game, The Dealer's name is really important thing? 🤔 
Maybe, it just need only for the view layer!

src/main/kotlin/blackjack/controller/GameLogic.kt
```diff
@@ -0,0 +1,21 @@
+package blackjack.controller
+
+import blackjack.model.GameResult
+
+object GameLogic {
+
+    fun getGameResult(
```
Hyeon9mak, line 7: > when not building an app with clear API endpoints (like a web controller), some recommend using a service layer instead of a controller directory to hold the logic (e.g., GameController). Would that be more appropriate in this case?

In the end, `Controller` and `Service` are just terms that express the idea of orchestration.
What really matters is having a solid bridge between the `Model`(`Domain`), which contains the core Blackjack rules, and the View, which handles input and output.
So honestly, any naming could work.

However, I’d like to offer one suggestion.
Many developers tend to use names like
`Controller`, `Service`, `Manager`, `Utils`, `Factory` out of habit.
As you can guess, these names often end up representing large, catch-all classes that can do anything and hold anything.
This makes it hard to guess what logic is inside without checking, causing inconvenience and increasing the risk of merge conflicts, which hurts maintainability.

Therefore, I encourage practicing with more specific and intuitive names instead of generic ones like `Controller` or `Service`.
For example… how about something like `BlackjackGame`?

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,29 @@
...
+
+        while (total > 21 && aces > 0) {
+            total -= 10
+            aces--
+        }
+
+        return total
+    }
+
+    fun isBusted(): Boolean = getScore() > 21
```
Hyeon9mak, line 24: 😉 👍

src/test/kotlin/blackjack/model/HandTest.kt
```diff
@@ -0,0 +1,37 @@
...
+        assertEquals(18, hand.getScore())
+    }
+
+    @Test
+    fun `should detect hand as busted when score over 21`() {
+        hand.add(Card(Suit.HEART, Rank.KING))
+        hand.add(Card(Suit.CLUB, Rank.QUEEN))
+        hand.add(Card(Suit.SPADE, Rank.TWO))
+        assertTrue(hand.isBusted())
+    }
```
Hyeon9mak, line 36: 👍 And I think much better append the over 21 Ace test 😉

Hyeon9mak (approved):
Hi Sari!
I'm glad to took the feedback comments positively.

I’ve left a few additional comments and also answered your question.
You seem to have a solid grasp of everything — just a few minor tweaks needed. 😄

I’ll go ahead and move you to the next step. See you in step2!
If you have any more questions, feel free to leave a comment 😉

src/main/kotlin/blackjack/model/Dealer.kt
saragrosser: I've now removed all references to InputView and OutputView from Dealer and moved the related logic into TurnManager. The model no longer depends on any view layer, and the controller fully coordinates interactions between model and view.

Also passed the "Dealer" name explicitly into Participant to keep the display logic separate from the game logic 😄

Thank you again for the guidance!

src/main/kotlin/blackjack/view/SuitView.kt
saragrosser: I have refactor to make suit private and I also did rank private! Thank you!

src/main/kotlin/blackjack/model/GameResult.kt
saragrosser: Yes I understand now better, I have created a Gameresultview file now to separate the view from the logic! thank you!!

PR #12, opened 2025-07-10
[Blackjack - Step1] Submit Farhana's mission

farhanaahmed:
This PR implements the core logic of a simplified Blackjack game using the MVC architecture in Kotlin.
1. Is using an interface for Participant a good design decision here, or would an abstract class offer better code reuse in this case?
2. Does the Participants class help improve clarity and structure, or could it be simplified further?
3. Any suggestions for improving testability?

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,30 +5,31 @@ The goal is to get as close to 21 as possible without going over.
 
 ## Features 
 - Input View 
-- [x] ask user inputs for player name
-- [x] get list of names from the user input
+- [x] Ask user inputs for player name
+- [x] Get list of names from the user input
+- [x] Ask player's choice to draw card or stay 
 
-- Card 
-- [x] Visibility of the card implemented by flipCard() 
-- [x] Determine the value of each card
+- Card
 - [x] Create Enum class - Ranks
-- [x] Create Class - Card
 - [x] Create Enum class - Suits
+- [x] Create Class - Card
+- [x] Determine the value of each card
 
 - Deck
-- [x] prepare deck
-- [x] shuffle cards
+- [x] Prepare deck
+- [x] Shuffle cards
 
-- Participant interface 
-- [x] drawCard()
-- [x] update status
+- Participant : abstract class 
+- [x] Draw card
+- [x] Calculate total value of the cards
+- [x] Update status of a participant
 
 - Dealer
-- [x] use Participant interface
-- [x] implement additional methods to draw card
+- [x] Use Participant abstract class
+- [x] Implement additional methods to draw card
 
 - Player
-- [x] use Participant interface
+- [x] Use Participant abstract class
 
 - Game (controller)
 - [x] Initialize Game
--- a/README.md
+++ b/README.md
@@ -2,10 +2,8 @@
 
 ## Features 
 - Input View 
-- [ ] ask user inputs for player name
-- [ ] validate number of players (max 7 players)
-- [ ] validate name
-- [ ] get list of names from the user input
+- [x] ask user inputs for player name
+- [x] get list of names from the user input
 
 - Card 
 - [x] Visibility of the card implemented by flipCard() 
@@ -29,7 +27,7 @@
 - Player
 - [x] use Participant interface
 
-- Game 
+- Game (controller)
 - [ ] Initialize Game
 - [ ] Assign two cards to every member
 - [ ] Offer possibility to get a new card
@@ -37,11 +35,11 @@
 - [ ] Retrieve Winning Information
 
 - Result 
-- [ ] Get total value of the cards
-- [ ] Process Winning Information
+- [x] Get total value of the cards
+- [x] Process Winning Information
 
 - OutputView
-- [ ] Display cards
-- [ ] Display dealer message (if 16 or less)
-- [ ] Display all the cards with total value
-- [ ] Display Winning Result
+- [x] Display cards
+- [x] Display dealer message (if 16 or less)
+- [x] Display all the cards with total value
+- [x] Display Winning Result
--- a/README.md
+++ b/README.md
@@ -1,5 +1,8 @@
 # kotlin-blackjack
 
+A simple kotlin project implementation of the classic Blackjack card game where a player competes against a dealer. 
+The goal is to get as close to 21 as possible without going over.
+
 ## Features 
 - Input View 
 - [x] ask user inputs for player name
@@ -28,11 +31,11 @@
 - [x] use Participant interface
 
 - Game (controller)
-- [ ] Initialize Game
-- [ ] Assign two cards to every member
-- [ ] Offer possibility to get a new card
-- [ ] Check status
-- [ ] Retrieve Winning Information
+- [x] Initialize Game
+- [x] Assign two cards to every member
+- [x] Offer possibility to get a new card
+- [x] Check status
+- [x] Retrieve Winning Information
 
 - Result 
 - [x] Get total value of the cards
--- a/README.md
+++ b/README.md
@@ -1,1 +1,40 @@
 # kotlin-blackjack
+
+## Features 
+- Input View 
+[ ] ask user inputs for player name
+[ ] validate number of players (max 7 players)
+[ ] validate name 
+[ ] get list of names from the user input
+
+- Card 
+[ ] Create Enum class - Suits
+[ ] Create Emus class - Ranks
+[ ] Create Class - Card
+[ ] Visibility of the card - implemented by flipCard() 
+[ ] Determine the value of each card
+
+- Deck
+[ ] prepare deck
+[ ] shuffle cards
+
+- Member interface - drawCard(), update status
+[ ] Dealer - assign win and lose counter, 
+[ ] Players 
+
+- Game 
+[ ] Initialize Game
+[ ] Assign two cards to every member
+[ ] Offer possibility to get a new card
+[ ] Check status
+[ ] Retrieve Winning Information
+
+- Result 
+[ ] Get total value of the cards
+[ ] Process Winning Information
+
+- OutputView
+[ ] Display cards
+[ ] Display dealer message (if 16 or less)
+[ ] Display all the cards with total value
+[ ] Display Winning Result
--- a/README.md
+++ b/README.md
@@ -10,17 +10,24 @@
 - Card 
 - [x] Visibility of the card implemented by flipCard() 
 - [x] Determine the value of each card
-- [x] Create Emus class - Ranks
+- [x] Create Enum class - Ranks
 - [x] Create Class - Card
-
 - [x] Create Enum class - Suits
+
 - Deck
 - [x] prepare deck
 - [x] shuffle cards
 
-- Member interface - drawCard(), update status
-- [ ] Dealer - assign win and lose counter, 
-- [ ] Players 
+- Participant interface 
+- [x] drawCard()
+- [x] update status
+
+- Dealer
+- [x] use Participant interface
+- [x] implement additional methods to draw card
+
+- Player
+- [x] use Participant interface
 
 - Game 
 - [ ] Initialize Game
--- a/README.md
+++ b/README.md
@@ -2,39 +2,39 @@
 
 ## Features 
 - Input View 
-[ ] ask user inputs for player name
-[ ] validate number of players (max 7 players)
-[ ] validate name 
-[ ] get list of names from the user input
+- [ ] ask user inputs for player name
+- [ ] validate number of players (max 7 players)
+- [ ] validate name
+- [ ] get list of names from the user input
 
 - Card 
-[x] Create Enum class - Suits
-[x] Create Emus class - Ranks
-[x] Create Class - Card
-[x] Visibility of the card - implemented by flipCard() 
-[x] Determine the value of each card
+- [x] Visibility of the card implemented by flipCard() 
+- [x] Determine the value of each card
+- [x] Create Emus class - Ranks
+- [x] Create Class - Card
 
+- [x] Create Enum class - Suits
 - Deck
-[x] prepare deck
-[x] shuffle cards
+- [x] prepare deck
+- [x] shuffle cards
 
 - Member interface - drawCard(), update status
-[ ] Dealer - assign win and lose counter, 
-[ ] Players 
+- [ ] Dealer - assign win and lose counter, 
+- [ ] Players 
 
 - Game 
-[ ] Initialize Game
-[ ] Assign two cards to every member
-[ ] Offer possibility to get a new card
-[ ] Check status
-[ ] Retrieve Winning Information
+- [ ] Initialize Game
+- [ ] Assign two cards to every member
+- [ ] Offer possibility to get a new card
+- [ ] Check status
+- [ ] Retrieve Winning Information
 
 - Result 
-[ ] Get total value of the cards
-[ ] Process Winning Information
+- [ ] Get total value of the cards
+- [ ] Process Winning Information
 
 - OutputView
-[ ] Display cards
-[ ] Display dealer message (if 16 or less)
-[ ] Display all the cards with total value
-[ ] Display Winning Result
+- [ ] Display cards
+- [ ] Display dealer message (if 16 or less)
+- [ ] Display all the cards with total value
+- [ ] Display Winning Result
--- a/README.md
+++ b/README.md
@@ -8,15 +8,15 @@
 [ ] get list of names from the user input
 
 - Card 
-[ ] Create Enum class - Suits
-[ ] Create Emus class - Ranks
-[ ] Create Class - Card
-[ ] Visibility of the card - implemented by flipCard() 
-[ ] Determine the value of each card
+[x] Create Enum class - Suits
+[x] Create Emus class - Ranks
+[x] Create Class - Card
+[x] Visibility of the card - implemented by flipCard() 
+[x] Determine the value of each card
 
 - Deck
-[ ] prepare deck
-[ ] shuffle cards
+[x] prepare deck
+[x] shuffle cards
 
 - Member interface - drawCard(), update status
 [ ] Dealer - assign win and lose counter, 
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -2,63 +2,37 @@ package blackjack.controller
 
 import blackjack.model.Dealer
 import blackjack.model.Deck
+import blackjack.model.Participants
 import blackjack.model.Player
+import blackjack.model.Players
 import blackjack.view.InputView
 import blackjack.view.OutputView
 
-class Game(
-    private val deck: Deck,
-    private val dealer: Dealer,
-) {
+class Game {
+    private val deck = Deck()
+
     fun startGame() {
-        val playersName = InputView.askPlayerNames()
-        val players = playersName.map { Player(it) }
-        dealCardsToDealer()
-        dealCardsToPlayer(players)
-        OutputView.displayInitialCards(dealer, players)
-        askPlayersToDraw(players)
-        dealerDraws()
-        OutputView.displayCardsWithTotalValue(dealer, players)
-        compareFinalCards(players)
-        OutputView.displayFinalResults(dealer, players)
-    }
+        val names = InputView.askPlayerNames()
+        val players = Players(names.map { Player(it) })
+        val dealer = Dealer()
+        val participants = Participants(players, dealer)
 
-    fun dealCardsToDealer() {
-        dealer.drawCard(deck.draw(2))
-    }
+        participants.dealInitialCards(deck)
+        OutputView.showInitialCards(dealer, players.toList())
 
-    fun dealCardsToPlayer(players: List<Player>) {
-        players.forEach {
-            it.drawCard(deck.draw(2))
-        }
-    }
-
-    fun askPlayersToDraw(players: List<Player>) {
-        players.forEach {
-            do {
-                val answer = InputView.askToDrawCard(it.name)
-                if (answer) {
-                    it.drawCard(deck.draw(1))
-                    it.updateActiveStatus(it.calculateTotalValueOfCards())
-                    OutputView.displayHandCards(it)
-                }
-            } while (answer && it.isActive)
-        }
-    }
-
-    fun dealerDraws() {
-        val mustDraw = dealer.mustDraw(dealer.calculateTotalValueOfCards())
-        if (mustDraw) {
-            OutputView.displayDealerDrawMessage()
-            dealer.drawCard(deck.draw())
-        }
-    }
-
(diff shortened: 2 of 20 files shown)


src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,28 @@
+package blackjack.view
+
+class InputView {
```
pkeugine:
```suggestion
object InputView {
```

We can make the view classes as objects.
What do you think are the differences between a class and an object?
Official Kotlin documentation can be a great source to find this out :)

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,28 @@
...
+            throw IllegalArgumentException("Input can not be empty.")
+        }
+    }
+
+    fun askToDrawCard(name: String): Boolean {
+        println("Would $name like to draw another card? (y for yes, n for no)")
+        val choice = readln()
+        return when (choice) {
+            "y", "Y" -> true
+            "n", "N" -> false
```
pkeugine, line 24: Nice job on considering upper case inputs as well 👍

README.md
```diff
@@ -1 +1,48 @@
...
+- Input View 
+- [x] ask user inputs for player name
+- [x] get list of names from the user input
+
+- Card 
+- [x] Visibility of the card implemented by flipCard() 
+- [x] Determine the value of each card
+- [x] Create Enum class - Ranks
+- [x] Create Class - Card
+- [x] Create Enum class - Suits
```
pkeugine: Very clean README and commit logs 💯 
It was very easy to follow through the requirements and workflow of this PR.

src/main/kotlin/blackjack/model/Suit.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model
+
+enum class Suit(val symbol: Char) {
+    HEARTS('♥'),
+    DIAMONDS('♦'),
+    CLUBS('♣'),
+    SPADES('♠'),
```
But here's my thought:

Since the visual effects are used in console based output, I think they belong to the OutputView.

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
+package blackjack.model
+
+class Dealer(
+    override val name: String = "Dealer",
+    override var isActive: Boolean = true,
+    override val cardsInHand: MutableList<Card> = mutableListOf(),
+) : Participant {
+    fun mustDraw(totalValueOfCards: Int) = totalValueOfCards <= 16
```
pkeugine: I think this 16 could have a name since it's a magic number.

src/test/kotlin/blackjack/model/PlayerTest.kt
```diff
@@ -0,0 +1,45 @@
+package blackjack.model
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class PlayerTest {
+    val player = Player("Alex")
```
pkeugine: It's interesting to see various tests use the same `player`.
The tests get passed for now, but this kind of test fixture setup is not recommended
since in certain cases, tests will share the same fixture.

When tests share the same fixture, all the tests that use the fixture will get have strong coupling,resulting in unexpected test results.

Why don't we make tests use their own `Player` in order to prevent these cases?
Also, I want you to tell me why the tests get passed currently, and when this might fail.

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    fun askPlayersToDraw(players: List<Player>) {
+        players.forEach {
+            do {
+                val answer = inputView.askToDrawCard(it.name)
+                if (answer) {
+                    it.drawCard(deck.giveCard())
+                    it.updateActiveStatus(it.calculateTotalValueOfCards())
```
pkeugine: Since these logics are all about `Player`, why don't we keep it simple and let `Player` do all these with one simple call. :)

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model
+
+interface Participant {
+    val name: String
+    var isActive: Boolean
+    val cardsInHand: MutableList<Card>
```
pkeugine: Since `Participant` can evaluate total value of cards by itself,
I think `var isActive` is unnecessary since we can find out if they are active or not right away.
By having this `var isActive`, some other developer might change this by accident, creating a bug.
I'd like to know how you think about this :)

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    fun compareFinalCards(players: List<Player>) {
+        players.forEach {
+            if (it.calculateTotalValueOfCards() < dealer.calculateTotalValueOfCards()) {
+                it.isActive = false
+            }
+        }
+    }
```
pkeugine: This part can also be handled in `Player`.
By doing this, we'll be able to test this logic as well. 👍

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    fun getCards(): List<Card> {
+        return cards
+    }
+
+    fun giveCard(): Card {
+        require(cards.isNotEmpty()) { "Deck is empty!" }
+        return cards.removeAt(0)
+    }
+}
```
pkeugine: Instead of giving a single card at a time, having a parameter that tells `Deck` the amount of cards needed would greatly improve this method's usage.

Nice validation on emptiness though! 👍

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -0,0 +1,70 @@
...
+        outputView.displayCardsWithTotalValue(dealer, players)
+        compareFinalCards(players)
+        outputView.displayFinalResults(dealer, players)
+    }
+
+    fun dealCardsToDealer() {
+        repeat(2) {
+            dealer.drawCard(deck.giveCard())
+        }
+    }
```
pkeugine:
```suggestion
    private fun dealCardsToDealer() {
        repeat(2) {
            dealer.drawCard(deck.giveCard())
        }
    }
```

Let's keep things private if they aren't needed in outside world 🤫 .
Please check the whole project about this part.

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    fun displayDealerInitialCards(dealer: Dealer) {
+        println("Dealer: ${dealer.cardsInHand[0].rank.face}${dealer.cardsInHand[0].suit.symbol}")
+    }
+
+    fun displayHandCards(player: Player) {
+        println(player.storePlayerHand())
+    }
```
pkeugine: I'd like to know what your personal (or team) convention is about the order of methods in a class.
Ordering with specific intention and keeping it across the whole project really helps the maintainability. If you haven't thought of one, it'll be nice to think about it in this step!

src/test/kotlin/blackjack/model/CardTest.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.junit.jupiter.api.Test
+
+class CardTest {
+    @Test
+    fun `should create a card with given rank and suit`() {
+        val card = Card(Rank.TWO, Suit.HEARTS)
+        assertThat(card.rank.toString()).isEqualTo("TWO")
+        assertThat(card.suit.toString()).isEqualTo("HEARTS")
+    }
+}
```
pkeugine:
```suggestion
    @Test
    fun `should create a card with given rank and suit`() {
        val card = Card(Rank.TWO, Suit.HEARTS)
        assertThat(card.rank).isEqualTo(Rank.TWO)
        assertThat(card.suit).isEqualTo(Suit.HEARTS)
    }
}
```

Although it might seem not important, I think it would be better to use the enum itself
when asserting the contents inside a card.

I recommend using `toString()` only for logging or debugging purposes,
rather than as making an output, as part of a logic, or as test validation value.

Well, I guess `TWO` and `HEARTS` could count as debugging / logging info as well. 😂

src/test/kotlin/blackjack/model/DeckTest.kt
```diff
@@ -0,0 +1,25 @@
...
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class DeckTest {
+    val deck = Deck()
+
+    @Test
+    fun `deck should contain 52 cards`() {
+        assertThat(deck.getCards().count()).isEqualTo(52)
```
pkeugine:
```suggestion
        assertThat(deck.count()).isEqualTo(52)
```

Why don't we let deck count the cards by itself? 
Let's try our best to take out the value inside of a class only for output purposes. :)

src/test/kotlin/blackjack/model/PlayerTest.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    @Test
+    fun `should calculate sum of values of cards in hand`() {
+        val card1 = Card(Rank.TWO, Suit.HEARTS)
+        val card2 = Card(Rank.THREE, Suit.HEARTS)
+        player.drawCard(card1)
+        player.drawCard(card2)
+        assertThat(player.calculateTotalValueOfCards()).isEqualTo(5)
+    }
```
pkeugine: It would be great to cover more cases of different kinds of cards.
This will allow us to refactor with more confidence.

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    fun drawCard(card: Card) {
+        cardsInHand += card
+    }
+
+    private fun checkAces() = cardsInHand.count { card -> card.rank == Rank.ACE }
+
+    fun calculateTotalValueOfCards(): Int {
+        var totalValueOfCards = cardsInHand.sumOf { card -> card.rank.value }
+        var aceCounter = checkAces()
```
pkeugine: Just out of curiosity,
how come `totalValueOfCards` logic wasn't extracted into it's own method
while `checkAces` got one?

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,28 @@
...
+            else -> throw IllegalArgumentException("Input can only contain y for yes, n for no")
```
pkeugine: Screenshot 2025-07-11 at 2 35 36 PM

It would be great to forgive an invalid input! 😂

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model
+
+interface Participant {
```
pkeugine: **Q. Is using an interface for Participant a good design decision here, or would an abstract class offer better code reuse in this case?**

What do you think the differences are?
Is there a reason why you and your pair chose interface?

pkeugine: **Q. Does the Participants class help improve clarity and structure, or could it be simplified further?**

I think it does help improve the clarity and structure.
However, I'm curious about that further improvement. What is it like?

src/main/kotlin/blackjack/controller/Game.kt
pkeugine: **Q. Any suggestions for improving testability?**

I think some comments that I've written will actually help improve the testability.
Also, the comment about adding some more cases about cards in hand will also improve the effectiveness as well 👍

pkeugine (changes requested):
Hello Farhana! 🙌 
Nice work on blackjack mission, it was easy to read and follow 👍 

I've added some comments so be sure to check them out.
There are questions I included that I think will help you maintain this project.
Once you're done, hit me with another review request.

If you have any questions, feel free to leave comments here or ask via DM.
Happy hacking! 🎲 🎲 🎲

src/main/kotlin/blackjack/view/InputView.kt
farhanaahmed: Thanks for your nice suggestion.
I think class is used to create multiple instances with different states. One must instantiate it manually. An object is created as a single instance automatically and used for utilities or shared logic.
Since I’m using `InputView` and `OutputView` without needing multiple instances, defining them as `object` is better.

src/main/kotlin/blackjack/model/Suit.kt
farhanaahmed: I completely agree. Placing visual effects and card display inside `OutputView` makes a lot of sense since it's tied to the console output strategy.
That way, if we ever change the UI (like switching to a web interface or GUI), we can easily replace `OutputView` to another implementation without touching the game logic. It keeps our architecture flexible. I have removed visual logics from enum classes and added a method in `OutputView` to show these symbols.

src/main/kotlin/blackjack/model/Dealer.kt
farhanaahmed: Yes, you are right. My pair and I wanted to finish the program first and then refactor the magic numbers later. But in the end, we didn’t have enough time to refactor our code.In the new code, I have assigned a name to it and declare it as a companion object.

src/test/kotlin/blackjack/model/PlayerTest.kt
farhanaahmed: You're absolutely right. Using shared fixtures can lead to tightly coupled tests and unpredictable behavior. 😅

Right now, the tests are passing because:
Each test method creates and operates on the same `Player` reference, but none of them modify or depend on shared state across tests (e.g., no static/global object or reused instance between tests).

But it can fail if:
A test modifies the shared `Player`’s state (like adding cards) and another test assumes a clean slate.

To avoid these risks and make the tests more robust, I have updated the tests so each one creates its own new `Player` instance. That way, tests are completely isolated and safe from side effects.

Thanks for bringing this up, very helpful! 🙏

src/main/kotlin/blackjack/controller/Game.kt
farhanaahmed: You're right. Both `drawCard()` and `updateActiveStatus()` are tightly coupled with the player's own behavior. Wrapping them in a single method like `drawAndUpdate(deck)` improves encapsulation and makes the calling code cleaner. 🙌
I'll go ahead and refactor this into a method on `Player`. Thanks for the tip. :D

src/main/kotlin/blackjack/model/Participant.kt
farhanaahmed: Thanks for this insight. I agree that having a mutable `isActive` property can introduce the risks of accidental state changes and bugs.

Since we can determine whether a participant is active by evaluating their current total, it makes sense to remove the `isActive` variable and instead provide a function like `isStillInGame()` that calculates this.

This approach improves immutability and reduces the chance of inconsistent states. I have changed the code and added a function.

src/main/kotlin/blackjack/controller/Game.kt
farhanaahmed: Thanks for the suggestion. While moving the comparison logic into the `Player` class could improve encapsulation, in my current implementation I decided to keep the result evaluation inside `OutputView` for simplicity and centralized display logic.
This keeps the `Player` class focused only on maintaining its state, while `OutputView` handles both displaying cards and determining win/loss based on final totals.
I’ve also added extra test cases in `PlayerTest` to validate edge cases, especially around tricky Ace values and totals near 21, to ensure the scoring logic remains accurate

src/main/kotlin/blackjack/model/Deck.kt
farhanaahmed: Thanks for the feedback! I updated the giveCard() method to draw(count: Int = 1): List<Card>, allowing multiple cards to be drawn at once. This makes the deck API more flexible and helps simplify the game logic by reducing repetitive calls.I also added validation to ensure that deck has enough cards before drawing.

src/main/kotlin/blackjack/controller/Game.kt
farhanaahmed: Thanks for the suggestion. Making functions `private` when they’re not used externally is a great practice to encapsulate implementation details and reduce unintended usage.

I have reviewed the entire project and marked all such internal helper methods `private` to keep the API clean and focused.I hope that I didn't miss any.

src/main/kotlin/blackjack/view/OutputView.kt
farhanaahmed: Thanks for the great insights. I haven’t formally defined a method ordering convention yet, but I agree that consistency helps maintainability and readability across the project.

I’m thinking of adopting this general order for all classes:

1. Public methods first (the main API).
2. Then protected/internal methods.
3. Private helper methods at the bottom.
4. Group related methods together by functionality.
5. Keep lifecycle or entry-point methods near the top.

I’ll try to follow this convention throughout the project for better structure and easier navigation.

Would you recommend any other ordering rules or best practices you’ve found useful?

src/test/kotlin/blackjack/model/CardTest.kt
farhanaahmed: Thanks for the insightful suggestion! I agree that using the enum values directly in assertions is clearer and less fragile than relying on `toString()`. I have updated the tests to compare enums directly.

src/test/kotlin/blackjack/model/DeckTest.kt
farhanaahmed: I have added a `countCards()` method to `Deck` and updated tests to use it instead of accessing the internal card list directly. 
Thanks for the suggestion! 🙌

src/test/kotlin/blackjack/model/PlayerTest.kt
farhanaahmed: Yes, it would be great to test with different cases. I have added more unit tests for this class with different values.

src/main/kotlin/blackjack/model/Participant.kt
farhanaahmed: Good question! 😊
Initially, we extracted `checkAces` because it was a simple, reusable line, and we wanted to keep the main logic clean.
But I can see now the main `totalValueOfCards` logic is actually *more* complex and would benefit even more from being in its own method. I didn’t realize it at first. Fixed it now, and the code looks cleaner. Thanks.

src/main/kotlin/blackjack/view/InputView.kt
farhanaahmed: Haha 😄 good point. I agree that being a bit more forgiving would make the experience smoother for the player.
I have updated the logic to keep prompting the user until they enter a valid input instead of throwing an exception right away. Thanks for the suggestion! 🙌

src/main/kotlin/blackjack/model/Participant.kt
farhanaahmed: Great question. Both interface and abstract class can work here, but as far as I understood, there are some key differences : 

- **Interface**:
    - Defines a contract without state.
    - Allows classes to implement multiple interfaces (multiple inheritance of type).
    - Usually better for flexibility and multiple behaviors mixed in.
- **Abstract class**:
    - Can hold common state and concrete method implementations.
    - Enables code reuse by providing shared functionality in one place.
    - Only allows single inheritance.
  
In our case, since `Participant` has common properties (like `cardsInHand`) and shared methods (like calculating totals), an **abstract class** might actually be better to avoid duplicating code.
We chose an interface initially to keep things flexible and simple, but after seeing the shared logic, switching to an abstract class would improve code reuse and maintainability. I have updated my code and made `Participant` an abstract class. Please let me know your thoughts about this change. 🙏

farhanaahmed: Right now, `Participant` is doing *many* things:

- It stores all players and the dealer
- It handles dealing cards
- It controls player turns and dealer turns

But I had a thought to give some of its responsibilities to a different class :

- ✅ Keeping `Participant` only for storing and organizing players
- ✅ Moving turn-handling logic to another class like `ManageTurn`

But I am not sure if it’s a better idea or not. Looking forward to your valuable insights on this. 🙏

src/main/kotlin/blackjack/model/Suit.kt
pkeugine: PERFECTO EXPLANATION-O with examples such as GUI and web interface 💯

src/main/kotlin/blackjack/controller/Game.kt
pkeugine: Nice work on adding in extra test cases 💯 

Also, while I do respect and follow reviewee's opinion if their backing reasons are pretty good for most of the times, I think determining who wins and comparing players' & dealer's cards in OutputView is not a good strategy.

The reason is because just like we've talked about, `OutputView` that we're currently using is for console printing. Even if we change it to other output view strategy, we would still need that logic where we compare players' & dealer's cards and determine who wins.

However, I do understand that your implementation keeps `Player` class focus on maintaining state. Well, let's think about it while working on the next step :)

src/main/kotlin/blackjack/view/OutputView.kt
pkeugine: You're method actually seems perfect to me 😂

src/main/kotlin/blackjack/model/Participant.kt
pkeugine: It's great 👍 
This is also a common way of using an abstract class:
1. by making features first, 
2. then finding out what the shared logics are
3. and refactor using abstract class

Nice job!

pkeugine: I think it's actually fine as is. Having `ManageTurn` class would make it more difficult to manage,
since it really feels like we're trying to separate the logic because we `feel like` a class has too much responsibility.
But to me, since `Participants` has the list of players and dealers, it has the responsibility to get the deck and deal cards to them.

This also applies to [your previous comment about `Player` having too much responsibility](https://github.com/herotechcourse/kotlin-blackjack/pull/12#discussion_r2203583781). When we say `single responsibility` in SRE, I don't think it really means one unit of responsibility, such as.. `having a collection`, `dealing cards`, etc. When the class has all the ingredients such as `list of players` and `a dealer` in it, all the things related to them would be it's single responsibility.

If you have any other thoughts, I'm always here 😆

pkeugine (approved):
Hello again, Farhana! 🙌
It really feels like I've left a lot of comments, but nice work going through them all.
I've added some replies and additional thoughts on our topics, so it'll be nice to check them out as you work on your next step.

If you have any unanswered questions or thoughts, DM me anytime.
And as always,
Happy Hacking 🚀 🚀 🚀

PR #13, opened 2025-07-10
[Blackjack - Step1] Submit Arekiu's mission

arekiu:
# Blackjack game implementation

## Features
- Create deck with 52 shuffled cards
- Use interface to create Dealer and Players (with custom names)
- Assign cards and count total value of cards in the hand
- Determine winners depending on their status as active in the game 

## Implementation notes
- My team reached the deadline, so the priority was to achieve the functionality in the game. Definitely I´m willing to improve different points, that I already see are not the best implementation:
    - Game() has methods that should belong to a separate Class
    - The use of magic numbers
    - Validation of input
    - Less use of "forEach"
    - Take more advantage of functional programming
- I would like to know if the interface was well implemented or if in this case, it´s better to use an Abstract class

Thank you so much for your time, and I hope our code is easy to follow!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,48 @@
 # kotlin-blackjack
+
+A simple kotlin project implementation of the classic Blackjack card game where a player competes against a dealer. 
+The goal is to get as close to 21 as possible without going over.
+
+## Features 
+- Input View 
+- [x] ask user inputs for player name
+- [x] get list of names from the user input
+
+- Card 
+- [x] Visibility of the card implemented by flipCard() 
+- [x] Determine the value of each card
+- [x] Create Enum class - Ranks
+- [x] Create Class - Card
+- [x] Create Enum class - Suits
+
+- Deck
+- [x] prepare deck
+- [x] shuffle cards
+
+- Participant interface 
+- [x] drawCard()
+- [x] update status
+
+- Dealer
+- [x] use Participant interface
+- [x] implement additional methods to draw card
+
+- Player
+- [x] use Participant interface
+
+- Game (controller)
+- [x] Initialize Game
+- [x] Assign two cards to every member
+- [x] Offer possibility to get a new card
+- [x] Check status
+- [x] Retrieve Winning Information
+
+- Result 
+- [x] Get total value of the cards
+- [x] Process Winning Information
+
+- OutputView
+- [x] Display cards
+- [x] Display dealer message (if 16 or less)
+- [x] Display all the cards with total value
+- [x] Display Winning Result
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -11,18 +11,28 @@ class Game(
     fun startGame() {
         val playersName = InputView.askPlayerNames()
         val players = playersName.map { Player(it) }
+
+        assignInitialCards(players)
+
+        hitOrStay(players)
+
+        compareFinalCards(players)
+    }
+
+    private fun assignInitialCards(players: List<Player>) {
         dealer.selfDrawInitialCards()
         dealer.dealInitialCardsToPlayers(players)
         OutputView.displayInitialCards(dealer, players)
+    }
+
+    private fun hitOrStay(players: List<Player>) {
         askPlayersToHit(players)
         dealerDraws()
         OutputView.displayCardsWithTotalValue(dealer)
         players.forEach { player -> OutputView.displayCardsWithTotalValue(player) }
-        compareFinalCards(players)
-        OutputView.displayFinalResults(dealer, players)
     }
 
-    fun askPlayersToHit(players: List<Player>) {
+    private fun askPlayersToHit(players: List<Player>) {
         players.forEach {
             do {
                 val answer = InputView.askToHit(it.name)
@@ -35,19 +45,20 @@ class Game(
         }
     }
 
-    fun dealerDraws() {
-        val mustDraw = dealer.mustDraw(dealer.cardsInHand.calculateTotalValueOfCards())
-        if (mustDraw) {
+    private fun dealerDraws() {
+        var mustDraw = dealer.mustDraw(dealer.cardsInHand.calculateTotalValueOfCards())
+        while (mustDraw) {
             OutputView.displayDealerDrawMessage(dealer)
             dealer.selfDrawCard()
+            mustDraw = dealer.mustDraw(dealer.cardsInHand.calculateTotalValueOfCards())
         }
     }
 
-    fun compareFinalCards(players: List<Player>) {
+    private fun compareFinalCards(players: List<Player>) {
+        val dealerPoints = dealer.cardsInHand.calculateTotalValueOfCards()
         players.forEach {
-            if (it.cardsInHand.calculateTotalValueOfCards() < dealer.cardsInHand.calculateTotalValueOfCards()) {
-                it.isBusted = true
-            }
+            it.comparePointsAgainstDealer(dealerPoints)
         }
+        OutputView.displayFinalResults(dealer, players)
     }
 }
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -38,7 +38,7 @@ class Game(
                 val answer = InputView.askToHit(it.name)
                 if (answer) {
                     dealer.dealCardToPlayer(it)
-                    it.updateBustedStatus()
+                    it.checkCardsValueLimit()
                     println(OutputView.showHandCards(it, false))
                 }
             } while (answer && !it.isBusted)
@@ -52,12 +52,15 @@ class Game(
             dealer.selfDrawCard()
             mustDraw = dealer.mustDraw(dealer.cardsInHand.calculateTotalValueOfCards())
         }
+        dealer.checkCardsValueLimit()
     }
 
     private fun compareFinalCards(players: List<Player>) {
-        val dealerPoints = dealer.cardsInHand.calculateTotalValueOfCards()
-        players.forEach {
-            it.comparePointsAgainstDealer(dealerPoints)
+        if (!dealer.isBusted) {
+            val dealerPoints = dealer.cardsInHand.calculateTotalValueOfCards()
+            players.forEach {
+                it.comparePointsAgainstDealer(dealerPoints)
+            }
         }
         OutputView.displayFinalResults(dealer, players)
     }
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -14,7 +14,7 @@ class Game(
         dealer.selfDrawInitialCards()
         dealer.dealInitialCardsToPlayers(players)
         OutputView.displayInitialCards(dealer, players)
-        askPlayersToDraw(players)
+        askPlayersToHit(players)
         dealerDraws()
         OutputView.displayCardsWithTotalValue(dealer)
         players.forEach { player -> OutputView.displayCardsWithTotalValue(player) }
@@ -22,16 +22,16 @@ class Game(
         OutputView.displayFinalResults(dealer, players)
     }
 
-    fun askPlayersToDraw(players: List<Player>) {
+    fun askPlayersToHit(players: List<Player>) {
         players.forEach {
             do {
-                val answer = InputView.askToDrawCard(it.name)
+                val answer = InputView.askToHit(it.name)
                 if (answer) {
                     dealer.dealCardToPlayer(it)
-                    it.updateActiveStatus(it.calculateTotalValueOfCards())
+                    it.updateBustedStatus(it.calculateTotalValueOfCards())
                     println(OutputView.showHandCards(it, false))
                 }
-            } while (answer && it.isActive)
+            } while (answer && !it.isBusted)
         }
     }
 
@@ -46,7 +46,7 @@ class Game(
     fun compareFinalCards(players: List<Player>) {
         players.forEach {
             if (it.calculateTotalValueOfCards() < dealer.calculateTotalValueOfCards()) {
-                it.isActive = false
+                it.isBusted = true
             }
         }
     }
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -28,7 +28,7 @@ class Game(
                 val answer = InputView.askToHit(it.name)
                 if (answer) {
                     dealer.dealCardToPlayer(it)
-                    it.updateBustedStatus(it.calculateTotalValueOfCards())
+                    it.updateBustedStatus(it.cardsInHand.calculateTotalValueOfCards())
                     println(OutputView.showHandCards(it, false))
                 }
             } while (answer && !it.isBusted)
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -1 +1,48 @@
...
+- [x] Create Enum class - Ranks
+- [x] Create Class - Card
+- [x] Create Enum class - Suits
+
+- Deck
+- [x] prepare deck
+- [x] shuffle cards
+
+- Participant interface 
+- [x] drawCard()
```
Hyeon9mak, line 23: 😉 👍

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -0,0 +1,70 @@
...
+        }
+    }
+
+    fun dealCardsToPlayer(players: List<Player>) {
+        repeat(2) {
+            players.forEach {
+                it.drawCard(deck.giveCard())
+            }
+        }
+    }
```
Hyeon9mak: Right now, it looks like the Dealer and Player objects are just exchanging properties by exposing them to each other.

How about letting the Dealer and Player interact with each other more directly instead?

This would reduce the amount of information exposed to the Controller, making the code easier to maintain — and it would likely make writing test code much easier too!

```kotlin
dealer.selfDrawInitCards()
```

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
+package blackjack.model
+
+class Dealer(
+    override val name: String = "Dealer",
```
Hyeon9mak: If the name "Dealer" needs to be changed to below image file, which class file will a fellow developer open first: `OutputView`-`Dealer`?

image

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
...
+    override var isActive: Boolean = true,
```
Hyeon9mak: What does `isActive` mean? 🤔 
I think it would be more intuitive to use actual blackjack rules!

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
...
+    override val cardsInHand: MutableList<Card> = mutableListOf(),
+) : Participant {
+    fun mustDraw(totalValueOfCards: Int) = totalValueOfCards <= 16
+
+    fun getFinalResultForDealer(players: List<Player>): String {
+        val lose = players.count { it.isActive }
+        val win = players.count { !it.isActive }
+        return "Dealer: $win Win $lose Lose"
+    }
```
Hyeon9mak: It looks like View's responsibility. How do u think?

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,19 @@
...
+            }.shuffled().toMutableList()
+
+    fun getCards(): List<Card> {
+        return cards
+    }
+
+    fun giveCard(): Card {
+        require(cards.isNotEmpty()) { "Deck is empty!" }
+        return cards.removeAt(0)
+    }
```
Hyeon9mak:
```suggestion
    fun giveCard(): Card {
        require(cards.isNotEmpty()) { "Deck is empty!" }
        return cards.removeFirst()
    }
```

😉

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,37 @@
+package blackjack.model
+
+interface Participant {
+    val name: String
+    var isActive: Boolean
+    val cardsInHand: MutableList<Card>
```
Hyeon9mak: What is the different between interface and abstract class?

src/main/kotlin/blackjack/model/Suit.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model
+
+enum class Suit(val symbol: Char) {
+    HEARTS('♥'),
+    DIAMONDS('♦'),
+    CLUBS('♣'),
+    SPADES('♠'),
+}
```
Hyeon9mak: When the Blackjack app convert to Web browser image game,

```kotlin
enum class SuitView(
    val suit: Suit,
    val symbol: String,
) {
    // ...
}
```

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    fun askToDrawCard(name: String): Boolean {
+        println("Would $name like to draw another card? (y for yes, n for no)")
+        val choice = readln()
+        return when (choice) {
+            "y", "Y" -> true
+            "n", "N" -> false
+            else -> throw IllegalArgumentException("Input can only contain y for yes, n for no")
+        }
+    }
```
Hyeon9mak: "y", "n" strings are not important things of Blackjack game.

Or u can try this

```kotlin
package blackjack.view 

enum class UserAnswer(
    private val value: String,
) {
    YES(value = "y"),
    NO(value = "n"),
    ;

    companion object {
        fun findByValue(value: String): UserAnswer = entries.find { value.equals(other = it.value, ignoreCase = true) }
            ?: throw IllegalArgumentException("Invalid user answer: $value")
    }
}
```

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,58 @@
+package blackjack.view
+
+import blackjack.model.Dealer
+import blackjack.model.Player
+import kotlin.collections.joinToString
+
+class OutputView {
```
Hyeon9mak: There is no property.
So u can change this class to object class!

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,37 @@
...
+            aceCounter--
+        }
+        return totalValueOfCards
+    }
+
+    fun updateActiveStatus(totalValueOfCards: Int) {
+        if (totalValueOfCards > 21) {
+            isActive = false
+        }
+    }
```
Hyeon9mak: I think u can give a new name to `21` score 😄

Hyeon9mak (changes requested):
Hello Alex!
U did a good job to complete Blackjack mission step1 👍

It looks like time was a bit tight, but I can see you made an effort to follow the basic MVC pattern.
It would be great to polish up the MVC structure a bit more.

Also, I noticed there are many parts in the code that could be replaced with actual blackjack game rules and terminology from the real world.
I think using real-world concepts would make the code much easier for all developers to understand. 
It would be awesome if you could give this some thought! 😄

src/main/kotlin/blackjack/model/Participant.kt
arekiu: I have been thinking about it. 
Interface is better when I want to define a role (behaviour by methods). Also if the class is going to have different roles, it could inherit from different interfaces. 
Abstract class is better to create classes that share properties and methods. So I think that it´s the best option for my project. 

Thank you for the question to making me reflect!

src/main/kotlin/blackjack/model/Deck.kt
arekiu: Thank you, I didn´t know that method!

src/main/kotlin/blackjack/view/OutputView.kt
arekiu: That´s right! I did the same with InputView

src/main/kotlin/blackjack/model/Dealer.kt
arekiu: If I understood right your question, you mean for example this:

`fun displayDealerDrawMessage() {
        println("Dealer draws one more card due to having 16 or less.")
    }`
    
   because I used the word "Dealer" multiple times in OutputView and then if I modify the name (like "CardGiver" or "Groupier") I need to make multiple changes in the file.
   
   I replaced for this:
   
   `fun displayDealerDrawMessage(dealer: Dealer) {
        println("${dealer.name} draws one more card due to having 16 or less.\n")
    }`
    
Please let me know in case I misunderstood the request.

src/main/kotlin/blackjack/model/Dealer.kt
arekiu: Absolutely it belongs to OutputView

src/main/kotlin/blackjack/controller/Game.kt
arekiu: Great idea! In order to improve the Game logic, I added the Deck as Dealer private property.

Before changing this logic, Players were accessing directly to the Deck and the responsibilities of Dealer were not clear.

src/main/kotlin/blackjack/model/Dealer.kt
arekiu: Clear! Replaced for "isBusted"

src/main/kotlin/blackjack/model/Participant.kt
arekiu: I created an object to store Constant values, now that 21 is "HAND_VALUE_LIMIT"

src/main/kotlin/blackjack/model/Suit.kt
arekiu: Oh I never thought about it. But it has sense because, for the logic of the game, I don´t need to know the symbols and face at all.
I created another enum class for the FaceView as well.

src/main/kotlin/blackjack/view/InputView.kt
arekiu: I appreciate a lot this suggestion and I was analyzing about the benefits of implement it or not.
Finally, I consider that it would add more complexity in this case where the input is simple, and as you mention not an important part of the game. I have the feeling that currently, that block using "when" is readable and simple.
 But definitively, I would apply this change if the functionality of the game depends on different options.

Of course, if you ask me to implement the change, I´m open and willing to do it!

arekiu:
Hello GuMak!

I implemented all the requested changes (except the one in Inputview but I left my comment there for your consideration).
I repeat, if you consider that I should apply that change, please let me know, because my wish is to learn about better practices!

Also, I refactored extra things to make the project cleaner:
- Created an object to store constant values
- Create wrapper class for Cards
- Increased test coverage
- Divided methods depending on Class responsibility

Thank you so much for your time and your attention to review my code.
Have a nice day!

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -0,0 +1,67 @@
...
+        dealer.dealInitialCardsToPlayers(players)
+        OutputView.displayInitialCards(dealer, players)
+    }
+
+    private fun hitOrStay(players: List<Player>) {
+        askPlayersToHit(players)
+        dealerDraws()
+        OutputView.displayCardsWithTotalValue(dealer)
+        players.forEach { player -> OutputView.displayCardsWithTotalValue(player) }
+    }
```
Hyeon9mak, line 33: 😉 👍

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,27 @@
+package blackjack.model
+
+class Dealer(name: String = "Dealer") : Participant(name) {
+    private val deck = Deck()
+
+    fun mustDraw(totalValueOfCards: Int) = totalValueOfCards < BlackJackValues.DEALER_STAND_CONDITION
+
+    fun selfDrawCard() {
```
Hyeon9mak, line 8: > Please let me know in case I misunderstood the request.

Oh, what I meant was "Really need to give the dealer a name?"
In a Blackjack, the dealer's name doesn't seem to be an important part of the business logic.
It's more of a UI-level distinction between the player and the dealer.

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -0,0 +1,67 @@
...
+        players.forEach {
+            do {
+                val answer = InputView.askToHit(it.name)
+                if (answer) {
+                    dealer.dealCardToPlayer(it)
+                    it.checkCardsValueLimit()
+                    println(OutputView.showHandCards(it, false))
+                }
+            } while (answer && !it.isBusted)
+        }
```
Hyeon9mak, line 45: And I think u can remove the `!` operator by negative question method.

```kotlin
while (answer && it.isNotBusted)
```

src/main/kotlin/blackjack/model/Card.kt
```diff
@@ -0,0 +1,4 @@
+package blackjack.model
+
+class Card(val rank: Rank, val suit: Suit) {
+}
```
Hyeon9mak, line 4: Question.
one `♦️A` card and another `♦️A` card on the table.

Should we treat these as the same card or as two distinct cards?

That decision could help determine whether we should use a data class or a regular class here.

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,23 @@
...
+        private set
+
+    fun drawCard(card: Card) {
+        cardsInHand.addCard(card)
+    }
+
+    fun updateBustedStatus() {
+            isBusted = true
+    }
+
```
Hyeon9mak, line 17: Managing the busted status as a property is a valid approach,
but we could also consider asking the Hand in real-time whenever we need that information.

```kotlin
cardInHand.isBusted()
```

src/main/kotlin/blackjack/view/RankView.kt
```diff
@@ -0,0 +1,23 @@
...
+    FIVE(Rank.FIVE, "5"),
+    SIX(Rank.SIX, "6"),
+    SEVEN(Rank.SEVEN, "7"),
+    EIGHT(Rank.EIGHT, "8"),
+    NINE(Rank.NINE, "9"),
+    TEN(Rank.TEN, "10"),
+    JACK(Rank.JACK, "J"),
+    QUEEN(Rank.QUEEN, "Q"),
+    KING(Rank.KING, "K"),
+    ACE(Rank.ACE, "A");
```
Hyeon9mak, line 18: 😉

Hyeon9mak (approved):
Hello Alex!
Great job the reflect the feedback.

I’ve left a few minor comments.
feel free to keep them in mind as you move on to the step2.
And if you have any questions, don’t hesitate to ask!

Hope u have a nice weekend!

src/main/kotlin/blackjack/model/Card.kt
arekiu: I understand. As class they would be different because their equality is based in reference. But as data class, their equality is based in their attributes. So it´s better for Card, cause they just store immutable values

src/main/kotlin/blackjack/model/Dealer.kt
arekiu: Now I understand your point! I could create the Dealer without assigning the name.
Anyway, I would keep the field because if for some reason, someone wants to be the Dealer, I can replace the name while creating the Object.
In the other hand, for myself, is more visually clear when I see $dealer.name in a string, indicating from who is coming that action. I guess another option would be creating a method to show the name of the class itself?

src/main/kotlin/blackjack/model/Participant.kt
arekiu: Ok clear!
I added this method in Cards class: fun checkIfIsBustHand()

src/main/kotlin/blackjack/controller/Game.kt
arekiu: Now I replaced isBusted for isPlaying in Participant class. I think is more clear because a player could go out of the game if they are busted but also if they have less points than dealer.

PR #14, opened 2025-07-10
[blackjack - Step1] Submit vittorio's mission

VittorioDeMarzi:
Pull Request: Step1 Implementation of BlackJack Game

Hi 🙂

Thanks in advance for taking the time to review our code and share your feedback. This pull request contain our implementation to the BlackJack game project. 

What wer did:
- Implemented the basic logic for a multiplayer BlackJack game.
- The game supports multiple players and a dealer, with separate rounds for each.
- Introduced a Statistics class to calculate the final results and print a summary.
- The GameManager handles the game setup and main loop.

What could be improved
- Currently, both Player and Dealer use the same class. Their behavior is distinguished using conditions.
- I'm considering introducing a common abstract class or interface (e.g., Participant) to separate responsibilities and avoid branching logic.

Questions I'd appreciate feedback on
- Do you think would it make more sense to extract a common interface or use an abstract class to share logic between them?
- Do you think the project structure and the separation into model, view, and controller packages follows the MVC pattern well, or are there improvements you’d suggest?
- Does GameManager take on too many responsibilities (setup, game logic, handling player vs dealer turns)?
- Is this method clear and maintainable? Or does the dealer vs player logic feel a bit forced?
- I’m currently passing InputView.askForCard() as a lambda to decouple input from game logic. Do you think this is a clean approach?

Looking forward to your thoughts and suggestions! 😊


src/main/kotlin/blackjack/model/GameManager.kt
```diff
@@ -0,0 +1,64 @@
+package blackjack.model
+
+import blackjack.view.OutputView
+
+class GameManager(private val dealer: Player, private val players: List<Player>) {
+    private val cardDeck = CardDeck()
+
+    fun setUp() {
```
Wordbe: If something must always be executed when GameManager is created, it might be a good idea to put it inside an init block! 😊

src/main/kotlin/blackjack/controller/BlackJackController.kt
```diff
@@ -0,0 +1,28 @@
...
+            val names = InputView.getPlayersName()
+            val players = PlayerFactory.with(names)
+            val dealer = PlayerFactory.createDealer()
+
+            val gameManager = GameManager(dealer, players)
+            gameManager.setUp()
+            OutputView.printAllPlayers(listOf(dealer) + players)
+            gameManager.playGame(dealer, players) { InputView.askForCard() }
+
+            val statistics = Statistics(dealer, players)
```
Wordbe: It would be better to declare it on the line right after the code where it is used, keeping it close to where it's needed. 🙂

src/main/kotlin/blackjack/model/Card.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.model
+
+data class Card(val suit: Suit, val rank: Rank) {
+    override fun toString(): String {
```
Wordbe: This method seems to be intended for the view, but it looks like the domain is depending on the view. How about separating them? 😊

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun calculatePoints(): Int {
+        val countOfAce = hold.cards.filter { it.rank.value == 1 }.size
+        var sum =
+            hold.cards.sumOf {
+                it.rank.value
+            }
+
+        repeat(countOfAce) {
+            if (sum + 10 <= 21) sum += 10
```
Wordbe: Try defining the magic number as a constant!

src/main/kotlin/blackjack/model/Person.kt
```diff
@@ -0,0 +1,4 @@
+package blackjack.model
+
+@JvmInline
+value class Person(val name: String)
```
Wordbe: The name Person suggests that the class could have many attributes beyond just name, so it might not be suitable as a value class. How about renaming it to something more specific? 🙂

src/main/kotlin/blackjack/model/Hold.kt
```diff
@@ -0,0 +1,21 @@
+package blackjack.model
+
+class Hold(hold: Set<Card>) {
```
Wordbe: I'm curious what's the difference between Hand and Hold? 
If Hold is a value class wrapping a collection of Card objects, how about naming it Cards instead?

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,25 @@
...
+
+class Hand() {
+    private val hold: Hold = Hold(setOf())
+
+    val cards
+        get() = hold.cards
+
+    fun numberOfCards() = hold.cards.size
+
+    fun addCard(card: Card) = this.hold.add(card)
```
Wordbe, line 11: You could remove this here for consistency, since it's not used elsewhere. 😊

src/main/kotlin/blackjack/model/Statistics.kt
```diff
@@ -0,0 +1,41 @@
+package blackjack.model
+
+class Statistics(private val dealer: Player, private val players: List<Player>) {
+    val totalResult
```
Wordbe: You might not need to create a member variable here!

Wordbe (changes requested):
Hello, Vito!
Nice to meet you.

Thank you for your thoughtful work on the Blackjack mission. The structure of your code looks pretty solid.

Please check the review comments, reflect on the parts that are feasible, and feel free to ask questions if anything is unclear! 😊

---
Q1.
Do you think would it make more sense to extract a common interface or use an abstract class to share logic between them?

It would be helpful to distinguish based on the presence or absence of state! 

Q2. Do you think the project structure and the separation into model, view, and controller packages follows the MVC pattern well, or are there improvements you’d suggest?

Yes, overall it looks good. However, it would be better to decouple the domain from the view dependency. 

Q3. Does GameManager take on too many responsibilities (setup, game logic, handling player vs dealer turns)?

Not too much right now, but it might be worth considering if things get a bit more complex. 

Q4. 
Is this method clear and maintainable? Or does the dealer vs player logic feel a bit forced?

It seems better to separate the dealer and players rather than managing them as a single player entity! 

Q5. 
I’m currently passing InputView.askForCard() as a lambda to decouple input from game logic. Do you think this is a clean approach?

Yes, there seem to be pros and cons to this pattern, and personally, I prefer simpler structures, so I don’t use it often. It can vary depending on the individual or the team!

---

Thank you!

src/main/kotlin/blackjack/model/GameManager.kt
VittorioDeMarzi:
```
class GameManager(private val dealer: Player, private val players: List<Player>) {
    private val cardDeck = CardDeck()

    init {
        players.forEach { cardDeck.hit(it, 2) }
        cardDeck.hit(dealer)
    }
```

something like this?

src/main/kotlin/blackjack/controller/BlackJackController.kt
VittorioDeMarzi: That's a simple but effective suggestion 🙂

src/main/kotlin/blackjack/model/Hold.kt
VittorioDeMarzi: This is totally meaningful :D

src/main/kotlin/blackjack/model/Hand.kt
VittorioDeMarzi: I am actually using it to hit the card to the player 😌

VittorioDeMarzi:
Hi Wordbe :)

Thanks a lot for the review and all your helpful feedback!

I’ve already addressed some of the points you mentioned, but I’ve kept the unified Player class for now instead of separating Dealer and Player. Before I go further with that refactor, I’d like to ask for your input.

Regarding your comment:
“It would be helpful to distinguish based on the presence or absence of state!”
I completely agree, and today in class we actually practiced applying the State Pattern (polymorphism) to the Blackjack project. It made me think it could be really interesting to integrate that approach into this project.
However, doing so would mean discarding most of the current structure and starting over from scratch to rebuild it around the pattern. It might take me a bit more time, do you think it’s worth trying now, or would it be better to first move on to Step 2?

Lastly, I also added a small DSL builder just to practice what we learned during the session. 🙂

Thanks again for your support  ☺️

src/main/kotlin/blackjack/model/Hand.kt
Wordbe: I'm sorry I wasn't clear. I was referring to `this` 🙂

src/main/kotlin/blackjack/model/GameManager.kt
Wordbe: Yes exactly!

src/main/kotlin/blackjack/dsl/PersonBuilder.kt
```diff
@@ -0,0 +1,61 @@
+package blackjack.dsl
+
+data class Person(val name: String, val company: String?, val skills: Skills, val languages: Languages)
```
Wordbe, line 3: Great DSL 👍

src/main/kotlin/blackjack/model/Card.kt
```diff
@@ -0,0 +1,3 @@
+package blackjack.model
+
+data class Card(val suit: Suit, val rank: Rank)
```
Wordbe, line 3: Try limiting the number of Card instances that can be created infinitely. 🙂

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    private fun printWinOrLose(condition: Boolean): String {
+        return when (condition) {
+            true -> "Win"
+            false -> "Lose"
+        }
+    }
+
+    fun printStatistics(statistics: Statistics) {
+        println("\n## Final Results")
```
Wordbe, line 36: Combine multiple prints into a single string template and call println only once!

Wordbe (approved):
Hello Vito!
Since I believe you has reflected most of the feedback well in this step, let's go ahead and merge and wrap it up here. 😊

Here are a few suggestions for the next step:

- Please apply the remaining review comments I mentioned earlier at the beginning of the next step.

- As you mentioned, consider separating Player into Dealer and Player to reflect their distinct roles.

- Try managing each player's state using a State pattern.

Thank you for thoughtfully applying the reviews.
Great work!

src/main/kotlin/blackjack/model/Card.kt
VittorioDeMarzi: I implemented the patter object pool in the Car Class. 
I think this is the right way for 
- Memory optimization: There are just 52 card in memory. No matter how many Desks we create. 
- Performance: We don't need to create new instances every time.
- Identity (Card1 === Card2 if they are equal)

Thanks for this feedback. This was really thoughtful. We've seen the Pool Pattern in Lotto with Jason but I didn't implemented yet :)

PR #15, opened 2025-07-10
[blackjack - Step1] Submit miooo0o's mission

miooo0o:
## Top Priorities
- The act of giving and receiving cards, and the subject or result of that act
- Design it so that it is easy to refactor later
- Separate as much as possible the parts that can be called 'controller'
- Expose InputView as much as possible (if it is hidden, it won't be easy to find)
- Make it as "injectable" as possible to test

## Implementation Results

- Separated core concepts into distinct classes: `Card`, `Hold`, `CardDeck`, `Hand`, `Player`
- `Player` has `Hand`, `Hand` has `Hold.cards`, `CardDeck` has `Hold(52 cards)`
- Used `Hold` as an abstraction for "something that holds cards" (shared by CardDeck and Hand)
- Made `CardDeck` immutable (data class) vs `Hand` mutable (game state changes)

- `Player` was implemented last and kept simplest. The dealer and player use the same class, distinguished by the actual instance reference where the dealer was initialized. It's not an ideal approach, but I thought it was efficient at first.
- Didn't know how to implement the controller with TDD, so couldn't proceed with TDD after the core logic was completed. (I think I understand a little bit now)
- Statistics: Used Int (0/1) instead of enum/sealed class because the dealer needs to count results (dealer identified by reference... this also impacts a bad decision by the design)
- Ace logic: Chose simple iteration over complex optimal calculation for speed

## Current Limitations
- There is still hardcoded logic or magic numbers left
- Lambda methods do not look or use clean
- Some structural imperfections remain (like Player interface-like implementation)
- I'm not sure if each class is doing just one thing. After focusing on implementation, it seems like responsibilities got mixed up.
- Just skip over some annoying edge cases: What if the CardDeck is empty?

## Reflection
I tried not to find the best solution in this mission. There are parts that I didn't care about if they were somehow feasible, even if they weren't structurally perfect. I still think it's a trade-off.


README.md
```diff
@@ -1 +1,42 @@
...
+- Card values follow standard Blackjack rules:
+- Number cards are counted by their face value.
+- Face cards (King, Queen, Jack) are each worth 10.
+- Aces can be worth either 1 or 11.
+- Each player starts with two cards.
+- Players may draw additional cards as long as their total remains 21 or less.
+- The dealer must draw a card if their total is 16 or less, and must stand on 17 or more.
+- If the dealer busts (goes over 21), all remaining players automatically win.
+- After the game ends, display the result (win/loss) for each player.
+-
```
pkeugine: Nice job organizing README and commits 👍 
It's a very small detail, but I see a trailing `-` mark here which can be removed 👀

src/main/kotlin/blackjack/controller/BlackJackController.kt
```diff
@@ -0,0 +1,28 @@
...
+            val dealer = PlayerFactory.createDealer()
+
+            val gameManager = GameManager(dealer, players)
+            gameManager.setUp()
+            OutputView.printAllPlayers(listOf(dealer) + players)
+            gameManager.playGame(dealer, players) { InputView.askForCard() }
+
+            val statistics = Statistics(dealer, players)
+            OutputView.printFinalResults(listOf(dealer) + players)
+            OutputView.printStatistics(statistics)
```
pkeugine: If there's a consecutive `OutputView` call, why don't we merge them into one?

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,25 @@
+package blackjack.model
+
+class Hand() {
+    private val hold: Hold = Hold(setOf())
```
pkeugine:
```suggestion
    private val hold: Hold = Hold(emptySet())
```

What do you think is the difference between these two?
**Hint**: I recommend checking out the implementation of `setOf()` :)

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -0,0 +1,24 @@
...
+    val cards
+        get() = hand.cards
+
+    val isBlackJack
+        get() = numberInHand() == 2 && calculatePoints() == 21
+    val isBust
+        get() = calculatePoints() > 21
+
+    fun cardsToString() = cards.joinToString(", ")
+}
```
pkeugine: Nice job creating `Player` class and giving it the responsibility of calculating what is needed 💯 

To make it even better, it would be nice to
* make things private if it's only used inside `Player` (for example : `cards`)
* Explicitly show the type of each methods
* and order components so that methods stay together.

Also, why would explicitly declaring types be helpful?
I would like you to think about it :)

src/main/kotlin/blackjack/model/Hold.kt
```diff
@@ -0,0 +1,21 @@
...
+        to.addCard(target)
+    }
+
+    fun add(card: Card) {
+        _cards.add(card)
+    }
+
+    operator fun plus(other: Hold): Hold {
+        return Hold((other.cards + this.cards))
+    }
```
pkeugine: This logic doesn't seem like it's used anywhere.
Why don't we remove this for now and add it when it's necessary?

src/test/kotlin/blackjack/model/HandTest.kt
```diff
@@ -0,0 +1,60 @@
...
+                Card(Suit.DIAMONDS, Rank.TWO),
+                Card(Suit.DIAMONDS, Rank.QUEEN),
+                Card(Suit.DIAMONDS, Rank.FOUR),
+            )
+        listOfCards.forEach { player.addCard(it) }
+        assertThat(player.calculatePoints()).isEqualTo(17)
+    }
+
+    @Test
+    fun hasPoints2() {
```
pkeugine: I think it would be much better to manage tests and utilize them
if we named tests with more detail.

By naming them like `someTest1`, `someTest2`, `someTest3`
we won't be able to find out what it's actually asserting unless we read the test thoroughly.

You've done a great job on other tests, so let's keep up with it!

src/main/kotlin/blackjack/model/Hold.kt
```diff
@@ -0,0 +1,21 @@
...
+class Hold(hold: Set<Card>) {
+    private val _cards: MutableSet<Card> = hold.toMutableSet()
+    val cards: Set<Card>
+        get() = _cards.toSet()
+
+    fun moveCard(to: Player) {
+        val target = _cards.first()
+        _cards.remove(target)
+        to.addCard(target)
+    }
```
pkeugine: **PR message:** I'm not sure if each class is doing just one thing. After focusing on implementation, it seems like responsibilities got mixed up.

I actually think you did a pretty good job on this.
Again, great work on giving individual classes their own responsibilities.
This is just my opinion.

Instead of modifying the object (`Player`) that gets passed by parameter,
why don't we let `player.receiveCards` like : 
```Kotlin
val drawnCards = deck.drawCards(2)
player.addCards(drawnCards)
```

Passing players to deck in order to let them get more cards seems a little unintuitive.
By refactoring like above, we'll still be able to

```kotlin
        players.forEach { player ->
            player.addCards(cardDeck.drawCards(2))
        }
```
in `GameManager`.
Personally, I try not to modify objects that get passed by parameters since we call methods that hide implementations. Modifying values in hidden logic increases possibilities of unexpected behavior.

src/main/kotlin/blackjack/model/GameManager.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    private fun roundForDealer(player: Player) {
+        while (ableToReceive(player)) {
+            cardDeck.hit(player)
+        }
+        OutputView.printDealerDrawsCards(player)
+    }
+
+    private fun ableToReceive(player: Player): Boolean {
+        val isDealer = player === dealer
```
pkeugine: I see you've used `===` 👍 
I would like you to tell me the difference between `===`, `==`, and `equals()`.
If you know it already, let's think it as a review of what you've learned. :)

src/main/kotlin/blackjack/model/GameManager.kt
```diff
@@ -0,0 +1,64 @@
...
+        if (isDealer) return dealer.calculatePoints() <= ABLE_TO_RECEIVE
+        return player.calculatePoints() < BLACKJACK
+    }
```
pkeugine: I think we can let `Player` class handle this.
It'll be interesting to think about how we can recognize that `Player` object if it is a dealer or not :)

src/main/kotlin/blackjack/model/GameManager.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    companion object {
+        private const val ABLE_TO_RECEIVE = 16
```
pkeugine: I think this name could be more detailed since it's only applied to the dealer.

src/main/kotlin/blackjack/model/PlayerFactory.kt
```diff
@@ -0,0 +1,9 @@
+package blackjack.model
+
+object PlayerFactory {
+    fun with(names: List<String>) = names.map { Player(it) }
+
+    fun createDealer() = Player(DEALER)
+
+    const val DEALER = "Dealer"
```
pkeugine: Let's keep this as private since we only use this name in `PlayerFactory` class.
To think deeper, the `"Dealer"` string actually represents itself pretty well that
I don't think it needs a name that describes it. Why not just use the string itself?

src/main/kotlin/blackjack/model/Suit.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model
+
+enum class Suit(val symbol: String) {
+    HEART("♥"),
+    DIAMONDS("♦"),
+    CLUBS("♣"),
+    SPADES("♠"),
```
But here's my thought:

Since the visual effects are used in console based output, I think they belong to the OutputView.

src/main/kotlin/blackjack/model/PlayerFactory.kt
```diff
@@ -0,0 +1,9 @@
+package blackjack.model
+
+object PlayerFactory {
+    fun with(names: List<String>) = names.map { Player(it) }
+
+    fun createDealer() = Player(DEALER)
```
pkeugine:
```suggestion
    fun createPlayers(names: List<String>) = names.map { Player(it) }

    fun createDealer() = Player(DEALER)
```

I believe having consistent naming strategy allows us to maintain project more easily,
especially when working on it for a long time or with many teammates.

It makes code more predictable and less likely to make mistakes.
This is just my suggestion and I'm actually curious about your opinion :)

src/main/kotlin/blackjack/model/Person.kt
```diff
@@ -0,0 +1,4 @@
+package blackjack.model
+
+@JvmInline
+value class Person(val name: String)
```
pkeugine: This `@JvmInline` annotation seems interesting 😮 
I'm guessing you've added it because you wanted to use an inline value class.

That raises some questions:
* I'm curious about why you used an inline value class. What IS an `inline value class`?
* What does `@JvmInline` annotation do?

I'm not saying this is a wrong usage of inline value class. (actually it's great)
I'm just asking questions so that we get to know more about Kotlin .:)

src/main/kotlin/blackjack/model/GameManager.kt
pkeugine: **PR message:** The dealer and player use the same class, distinguished by the actual instance reference where the dealer was initialized. It's not an ideal approach, but I thought it was efficient at first.

We'll have to think about this again if we apply the above comment.

src/test/kotlin/blackjack/model/CardDeckTest.kt
```diff
@@ -0,0 +1,20 @@
+package blackjack.model
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class CardDeckTest {
```
pkeugine: **PR message:** Just skip over some annoying edge cases: What if the CardDeck is empty?

This is actually a really important case. I mean... WHAT HAPPENS IF THERE'S NO CARDS? `hit()` method should definitely be tested. It's the most important logic in `CardDeck`.

src/test/kotlin/blackjack/model/CardDeckTest.kt
```diff
@@ -0,0 +1,20 @@
...
+    fun `CardDeck has 52 cards`() {
+        val cardDeck = CardDeck()
+        assertThat(cardDeck.numberOfCards()).isEqualTo(52)
+    }
+
+    @Test
+    fun `hit a card to a player`() {
+        val cardDeck = CardDeck()
+        val player1 = Player("Mina")
+        cardDeck.hit(player1) // should move card to player
```
pkeugine: I think this comment best shows what I meant in the previous review about modifying objects that gets passed by parameter. By not letting it, I think we'll be able to remove this comment!

src/test/kotlin/blackjack/model/PlayerFactoryTest.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class PlayerFactoryTest {
+    @Test
+    fun with() {
+        val names = listOf("mina", "vito")
+        val players = PlayerFactory.with(names)
+        assertThat(players).hasSize(2)
+    }
```
pkeugine: Since `with()` method is about creating players **with names**, I think it's important that we assert names as well. 🙂

pkeugine (changes requested):
Hello Mina, nice work on blackjack mission 👍 
I see you've thought deeply about your work as you made it.
You've done a great job 💯 

I left some reviews and questions that I think will help you as you refactor the project.

If you have any questions or have any thoughts regarding those reviews,
feel free to leave comments in this PR or send me a DM anytime.
Happy hacking! ✈️ ✈️ ✈️

src/main/kotlin/blackjack/controller/BlackJackController.kt
miooo0o: It didn't get properly merged with them 😿

src/main/kotlin/blackjack/model/Hand.kt
miooo0o: I guess it's because 
1. it's more readable
2. it's inefficient to call emptySet() again in set Of()

I often forget empty when I'm typing fast. I'll pay more attention!

src/main/kotlin/blackjack/model/Player.kt
miooo0o: It doesn't look very readable. There definitely would have been ways to make it better. Thank you for advice!

src/main/kotlin/blackjack/model/Hold.kt
miooo0o: Yes, thank you for letting me know. I will review this again as much as possible and try to submit it cleanly.

src/main/kotlin/blackjack/model/PlayerFactory.kt
miooo0o: I thought about creating the dealer together with `with(names)`, but due to poor design choices -- checking the dealer by reference -- the dealer and player ended up being created separately with the same class. 

It would have been better to change it to "create", but I also liked the name "with" so I left it for the future. I thought this could also be applied to names later. but I can say now, was't that good idea 😢

src/main/kotlin/blackjack/model/Person.kt
miooo0o: In my understanding, 
`@JvmInline` is an annotation that tells the JVM how to handle value classes.

`Inline value class`es are classes that help remove the overhead of wrapper objects at compile time. They're useful because you can wrap primitives and use them safely. By distinguishing types at compile time, they don't create additional objects at runtime.

I think the way Kotlin operates is really fascinating, but I still don't fully understand the concept of how they "eliminate overhead."

src/test/kotlin/blackjack/model/CardDeckTest.kt
miooo0o: This is something I've been thinking about on and off for the past few days:

1. I used `set()` to create 52 unique cards for the Deck
2. I heard from Vito that real casinos use more cards (for example, they mix up to 8 decks into one pack)
3. Since I didn't have enough time to reconsider the `set()` approach, I postponed handling the empty case for later
4. I decided that in future refactoring, I'll create the deck as a List with 52 unique cards guaranteed, and add logic to regenerate it internally/externally

...and since `Hand/Deck` share logic, I wanted to merge them somehow while handling the `empty case` differently.

You'll see this in the PR I'm going to submit later.

miooo0o:
Hello, PK

Thank you for the feedback from last time.

In this PR, while applying the feedback you provided, I decided to implement one additional improvement that I felt was necessary or interested.

---

##  improvement
That improvement is **refactoring duplicate logic**. it wasn't as straightforward as I initially thought...

The reason I chose **refactoring duplicate logic** as an improvement point is as follows:

Since it was difficult to think through the details of how Blackjack would work from the beginning, Vito and I decided to think about what would be needed when playing the game in real life and implement it starting from the smallest units.

So we proceeded by branching out like this:

```
Card -> Hold -> CardDeck/Hand...
```

At this time, since we didn't consider inheritance, a structure was created where higher-level classes had lower-level classes as instances.

As a result, the final class had no choice but to go deep into internal properties to get `value`. It looked like `Player.cardsCount()`, but it was actually working like `Player.hand.getHold().getCards().size`.

Another problem that arose from this was that the logic felt separated even though it was combined.

Even though it looked like `CardDeck.hit(Player)`, internally it was calling `Hand.addCard()`, and inside that, `Hold.addCard()`... calling methods with the same name consecutively, which was sometimes confusing. Questions like: What methods am I actually using in `hit()`? In what order is this happening?

I divided my attempt to integrate **refactoring duplicate logic** into two approaches:

1. **Separation of Player/Dealer**: Clearly distinguish the roles and responsibilities of each to make them easy to use for classes that handle "actions"
2. **Organizing inheritance relationships**: Reorganize the separated classes into an understandable form

And the most important goal I considered was, refactoring itself. 

---

## Applied

The base for holding cards is now called **CardHolder**. CardHolder becomes Deck/Hand, and Hand becomes the base class for Participant, which is for all participants.

In reality, a "hand" cannot have a person, but in game logic, if participants don't have a "hand," they might as well "not exist." That's why I chose CardHolder as the starting point.

The actual implementation process was completely different. I didn't intend to merge these two groups. While creating Participant separately and CardHolder separately, I discovered a common structure and integrated them.

Through this structure, I was able to gather "**all card exchange logic**" into CardHolder, and specifically left the "behavior when `List<Card>` is Empty" that needs to work differently as an abstract function.

```
CardHolder (abstract)
├── Deck
└── Hand (open)
    └── Participant (abstract)
        ├── Player
        └── Dealer
```

Since I judged that there wasn't enough time to properly implement state management as well, I handled the `state` that Participant has as an enum class.

### Why did I choose abstract class?

There's no particular reason. I started with interface, but while fixing compiler errors, it became an abstract class. As I started gathering logic into `CardHolder`, it seemed more appropriate to provide functionality that works by default but operates differently only when needed.

### Other Changes

I added functionality for Player/Dealer to update State based on card count, Points, and round, and apply that to results. This is also a really important change, but it wasn't a part I specifically focused on in this PR. The functionality on the logic side has only been minimally tested.

---

## I'd especially liked to get feedback

### 1. **apply abstraction**

This was my first attempt at abstraction in Kotlin, so I'd like to hear feedback on my practice.

### 2. syntax with `override var`

Currently, Player has this structure:

```kotlin
override var _state: State = State.HIT
	get() = when {
		points > BLACKJACK_SCORE -> State.BUST
		isFirstRound() -> blackOrHit()
		else -> field
	}
```

I didn't want to use `override var`, especially, it is mutable, but still able to access from outside. but this was the best approach I could think of so far (To be honest, only this was working...)

Thank you again for taking the time to review this.

---

_Additionally, I accidentally pressed the button to call you - sorry about that._

src/main/kotlin/blackjack/model/PlayerFactory.kt
pkeugine: I wouldn’t say it was a bad idea and honestly, my suggestion wasn’t a big one either.
To put it simply: naming things consistently just helps. That’s all 🙂

src/main/kotlin/blackjack/model/Person.kt
pkeugine: This might be a hard topic for now, but the [official documentation's second sentence](https://kotlinlang.org/docs/inline-classes.html) gives us a hint.

src/test/kotlin/blackjack/model/CardDeckTest.kt
pkeugine: That makes sense about the Set vs List!
Just to clarify though, my concern was mostly about what happens at runtime if the deck runs out of cards.

What I intended in this review was to think about validating the deck size before we `hit`.
A nice `require` or `if statement` with exception of our choice, and a test case for it allows us to handle that edge case with desired result.

src/main/kotlin/blackjack/model/card/Suit.kt
```diff
@@ -0,0 +1,8 @@
+package blackjack.model.card
+
+enum class Suit(val symbol: String) {
+    HEART("♥"),
+    DIAMOND("♦"),
+    CLUB("♣"),
+    SPADE("♠"),
```
pkeugine, line 7: There are several reviews that haven't been answered or got applied in the project.

I'm not saying that all of the reviews must be applied, but I think that they should get some kind of reactions: answers if they are questions, or your opinion of why you don't think the suggestion is not a good fit for this project.

I think communication in PR is part of teamwork and soft skill. Refactoring is not only about what you (the PR creator) want to change, but is also about how a team manages the project :)

src/main/kotlin/blackjack/controller/GameManager.kt
```diff
@@ -0,0 +1,80 @@
...
+    private val participants: Participants,
+) {
+    private val deck = Deck()
+    private val dealer = participants.getDealer()
+    private val players = participants.getPlayers()
+
+    fun play(): GameResult {
+        setUp()
+        OutputView.showCards(dealer)
+        OutputView.showAllPayersCards(participants)
```
pkeugine, line 23: I didn't fully understand the [comment here](https://github.com/herotechcourse/kotlin-blackjack/pull/15#discussion_r2203058366).
I think we'll be able to merge consecutive `OutputView` calls into one.
Could you explain more about it? :)

README.md
```diff
@@ -1 +1,67 @@
...
+
+### Participant package
+
+```
+Hand -> Participant -> Player & Dealer -> Participants
+```
+
+**Think about this/check later**
+- [ ] Currently Participant has a rule that "all `Participant`s" have a "`name`", which is why it was changed to abstract class. Is this decision really necessary?
+- [ ] Delete/move methods for tests
```
pkeugine, line 50: Are these check boxes for the next step?
Also, since this document is for everyone who's related to this project,
it'll be nice to add more explanation about `Delete/move methods for tests`.
Just by reading that sentence, I'm not quite sure what it is.

src/main/kotlin/blackjack/controller/GameManager.kt
```diff
@@ -0,0 +1,80 @@
...
+import blackjack.model.participant.Dealer
+import blackjack.model.participant.Participant
+import blackjack.model.participant.Participants
+import blackjack.model.participant.Player
+import blackjack.model.result.GameResult
+import blackjack.model.state.State
+import blackjack.view.InputView
+import blackjack.view.OutputView
+
+class GameManager(
```
pkeugine, line 13: I'm curious about this design choice of having both `BlackJackGame` and `GameManager`,
since they both use classes related to view and model hence making them `a controller inside another controller`.

README.md
```diff
@@ -1 +1,67 @@
-# kotlin-blackjack
\ No newline at end of file
+# kotlin-blackjack
```
pkeugine, line 1: In the added PR comment about refactoring duplicate logic and your thinking process,
several things came up to my mind while reading it.
But first, thanks for explaining in details.

**1. about duplicate logic**
Personally, I don't think having duplicate logic is always bad. There are times where some parts look like they are duplicate so we refactor it, only to find out that it makes things harder to change or to implement new features. Worse, someone creates that duplicate logic again, since they didn't know the team refactored it. So I'm curious about why you thought refactoring duplicate logic was important in this project.

**2. duplicate logic and your confusion about which method was getting called**
To be honest, I'm not sure if I understood the relationship between these two. I can see that since there were many classes that got involved in every single feature, you wanted to keep things simple. Is my understanding correct? However, by looking at the code and `applied` section of the comment, I feel like the logic actually got way more complicated.

**3. `Player.cardsCount()` being `Player.hand.getHold().getCards().size`**
I don't see anything wrong with this as long as class separation makes sense and it feels necessary (All the classes having their respectable responsibilities such as validation). Also, if we want to keep it more object-oriented, we could make `cardsCount()` method to each classes and call it instead of taking out all the fields.

**4. final thought**
I'm surprised that you've managed to get all these work in this short time, but I think I'll have to mention this point about this PR:

Like I said in above comment, PR is about teamwork. You've posted a PR to get it merged, and comments are there to have some improvements or to fix bugs before it gets merged. Instead of focusing on the rework and asking for a feedback of your own interest, it'll be great to focus more about the review and sharing thoughts.

pkeugine (changes requested):
Hello Mina,
I go by the name PK here. 🙂 

Although we don't have much time,
I think there are points for you to go back and check before we move onto the next step.
I added some more reviews so please check them out.
I feel like they are very important parts of programming.

Happy Hacking! ⌨️ ⌨️ ⌨️

pkeugine:
Oh, no worries about pressing button, it's fine 👍

src/test/kotlin/blackjack/model/HandTest.kt
miooo0o: Oh, I didn't realize there were comments after this part. I thought there were fewer comments than I initially thought...

 It's particularly due to my capability limitation where I can't produce very fluent English writing at a fast pace.
I will try to write the test/name properly.

src/main/kotlin/blackjack/model/Hold.kt
miooo0o: Oh, why didn't I think of doing multiple draws? That's a really good idea.
I'll make sure to reflect it right away. That part was actually one of the things that really bothered me too.

src/main/kotlin/blackjack/model/GameManager.kt
miooo0o: I understood it as the difference between comparing actual objects versus comparing the values that objects hold.

I still think this choice wasn't too bad. When implementation needs to be fast, especially when I can't think anymore...

src/main/kotlin/blackjack/model/GameManager.kt
miooo0o: That's also a valid point. I don't know why, but I keep mistakenly thinking that the `player` also shares this logic.

what about 

```kotlin
 companion object {
        private const val DEALER_MUST_HIT_LIMIT = 16
```

src/main/kotlin/blackjack/model/PlayerFactory.kt
miooo0o: I had some thoughts about this. 

When considering whether `Player` and `Dealer `should share the same class, I felt it was more intuitive to set both as 'participants' in my mind. So I started with the rule that 'all participants have names' (even when I didn't implement `Participant/s` classes). The `Dealer` just operates differently from the `Player`, but they're both 'participants', so I thought it was right for the `Dealer` to have a name.

On the other hand, in the actual code, this additional memory usage based on my linguistic understanding isn't particularly useful. Still, I felt it was more intuitive for the dealer to follow the rule of 'having a name'. This isn't logical, it's just my intuitive feeling... I thought about it for a long time.

But since the logic still doesn't quite click for me, I'd like to keep it this way unless there's a 'specific reason to remove it'

src/main/kotlin/blackjack/model/Suit.kt
miooo0o: This is hard to answer clearly. I'm not sure what you're looking for.
Do you mean it would be better to have the `OutputView` figure out what this `Suit` represents?

If you could give me a hint about what other approaches there might be, it would be a great help to my understanding.

src/main/kotlin/blackjack/model/PlayerFactory.kt
miooo0o: I see. How to set up and accept language and words is important to me, so it's a part that I keep thinking about, which is why I wrote a bit lengthy. I'll try to think about it a bit more simply. Thank you.

src/test/kotlin/blackjack/model/CardDeckTest.kt
miooo0o: That's definitely a method I could consider. However, when the game needs to continue and the maximum number of players isn't fixed, I thought creating a new deck would be a more appropriate approach.

Thank you for the kind advice.

src/main/kotlin/blackjack/controller/GameManager.kt
miooo0o: This is hard to explain. Those two feel a bit different to me, so I keep them separated.
I even remembered like, I had merged some two(I mean what you reviewed) until your review.
I'm not exactly sure what make me feel those two is not same (or should stay like this)

Let me try combining them into one.
Then I think I'll be able to answer your question

src/main/kotlin/blackjack/model/card/Suit.kt
miooo0o: Again, I'm sorry. I didn't see that there was a review here. I agree with what you said. I'll make sure to follow the basics first. Thank you. This is very helpful advice.

README.md
miooo0o: this is an unprofessional mistake on my part. I'll reflect this in the current refactor.

src/main/kotlin/blackjack/controller/GameManager.kt
miooo0o: There is no justification for this design. There was an intended approach, but it wasn't implemented properly: Blackjack controller allows other secondary roles to be reassembled.

The current GameManager has too many responsibilities.

1. Game initialization (setUp)
2. Round management (round, playPlayerRound, playDealerRound)
3. State validation (canReceiveCard)
4. Result generation (returning GameResult)
5. Participant and deck management

```kotlin
    private val gameSetup = GameSetup()
    private val roundManager = RoundManager()
    private val stateManager = StateManager()
    private val resultCalculator = ResultCalculator()
```

This approach would have been much cleaner. Looking back now, I should have refactored this part more thoroughly instead of using inheritance.

README.md
miooo0o: 1. It was a simple reason. I was getting confused while writing the code, and I thought consolidating it in one place would make it easier.
_and... I wanted to revert it if I had more time, but I was already accustomed to the new logic, so I compromised with reality and set today as my PR deadline._

2. My intention was to consolidate card exchange logic in one class and have other classes inherit from it, then modify as needed. Did it  become that complicated? 

3. I see. I'll try to make it simpler next time.

4. Yes, I agree. I felt like I should ask questions after finished codes, but seem it make more difficulty.

Additionally, I could revert all the changes I've made so far, go back to simple step1, apply only your review feedback, and request your review again. That might not be a bad idea either. Because, up until a certain commits, there are only parts that I refactor codes with your reviews. I think it would be okay to start over from there. I could reorganize the reviews I found out about today and start from there and fix them.

As you said, the purpose of PR is communication and teamwork, so applying everything I thought would be good to try doesn't seem beneficial. 

What do you think?

From now on, I won't treat PR as part of my practice, but will approach it more realistically.

Thank you!

src/main/kotlin/blackjack/model/GameManager.kt
miooo0o: would it be better to leave it as is and apply it again in step2?
Or maybe I can just tear out that part from the new commits I made this time and apply it?

I wonder what do you think

src/test/kotlin/blackjack/model/HandTest.kt
pkeugine: It's fine. Just be sure to check before making a commit!
If that's too hard, even before making a PR would be fine.

src/main/kotlin/blackjack/model/GameManager.kt
pkeugine: Nice! Yes, you are correct and like I said before in the previous comment, I actually think you've done a great job 😄

src/main/kotlin/blackjack/model/GameManager.kt
pkeugine: It's fine. There's still tons of things we can talk about, and honestly, I think you got this in you already.
We can forget about it and focus on our current situation. :)

src/main/kotlin/blackjack/model/GameManager.kt
pkeugine: PERFECT 👍

src/test/kotlin/blackjack/model/CardDeckTest.kt
pkeugine: I do agree that creating a new deck could be a more appropriate approach.
However, that doesn't mean the `CardDeck` is free from validating it's own size before a `hit`.
It's always important for an object to know when things might go wrong, and throw an expected exception or behave like we told them to. I wouldn't say this is my own personal opinion, but rather a better way to handle exceptional cases. :)

src/main/kotlin/blackjack/model/Suit.kt
pkeugine: Yes. Instead of having "♥", "♦", "♣", "♠" in the `Suit`,
We can put them in OutputView since they are tightly coupled to console based printing strategy.

src/main/kotlin/blackjack/model/PlayerFactory.kt
pkeugine: I understand. 👍 
What I actually meant was a bit simpler point.
Since the string `"Dealer"` is already saying what it is,
we wouldn't need `const val DEALER = "Dealer"`, because that's like naming a car `"Car"`.

src/main/kotlin/blackjack/model/card/Suit.kt
pkeugine: It's fine! Getting these comments are the reason why we're doing this anyways.

src/main/kotlin/blackjack/controller/GameManager.kt
pkeugine: If you feel like they must be separated, I'm with you on that.

The reason why I feel like combining two methods that get consecutive calls is important,
is because it makes it easier for users of those api.

Having a single api is better than having multiple api since we need to think about the orders of which they get called, we'll need to read how to use them (having more api means having more documentation), and so on.

Having one, then separating when we need to is an easier strategy to follow.

src/main/kotlin/blackjack/controller/GameManager.kt
pkeugine: I think setup, rounds, result, and management could all be a one single responsibility : `Managing a game`.
Which perfectly fits the name `GameManager`.
Trying to separate responsibilities into micro level might actually make it harder to manage, ironically.
It's best to divide things when we really feel like we need to.

Besides, [I wrote this in my one of very first reviews.](https://github.com/herotechcourse/kotlin-blackjack/pull/15#discussion_r2199853385)
Your project looked very clean and easy to read when I saw it for the first time.

README.md
pkeugine: To be honest, I think your previous work, which is what I saw on the very first review request, was cleaner and easier to read and manage than what I see right now. I think you've put in a lot of thoughts in this project, which could've been what made me feel like it got harder. However, I still think it's a nice work. I wouldn't have been able to do all these if I were you 😮 

As for reverting, no, I don't think it's necessary. We'll go with this like you said, since we don't have much time to revert and think again. I'll go through your work on the next step since I think we've hit that point where we really need to get going 😂 

I think this was actually a pretty good time for us to think about PR and the review process :)
Thanks for checking out all the comments again, and leaving reactions to them.
I've also left my additional thoughts and reviews to them, so it'll be nice for you to check them out.

pkeugine: **As you said, the purpose of PR is communication and teamwork, so applying everything I thought would be good to try doesn't seem beneficial.**

This is your masterpiece, not mine. So I think it's fine.
Also, you've done an amazing job stacking up the commits. Let's keep them 👍

src/test/kotlin/blackjack/view/InputViewTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertThrows
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ValueSource
+
+class InputViewTest {
+    @Test
+    fun `name should be parsed correctly`() {
+        val input = "mina, guri, ella"
+        val parsed = InputView.Parser.playerNames(input)
```
pkeugine, line 14:
```suggestion
        val parsed = Parser.playerNames(input)
```

We can remove this :)

src/test/kotlin/blackjack/model/participant/PlayerTest.kt
```diff
@@ -0,0 +1,3 @@
+package blackjack.model.participant
+
+class PlayerTest
```
pkeugine, line 3: Let's delete this test class if we're not going to use it!

pkeugine (approved):
Hello Mina 🙂 
What a step it was huh?

I might've sounded a little harsh, but that's probably because I wanted to make sure I deliver some important points right away. But I'm amazed by how much work you've done, and how you've managed to check the reviews once again 👍 

We don't need to revert what you've already done, and we'll keep them as we move onto the next step. And don't say sorry, you're just doing your best, and I'm a person who suggests you the directions.

I didn't add too much reviews for now since we really need to move on,
but I've answered some comments.

If you find any of them not clear or have more questions, DM me anytime.

Again, nice work and as always,
Happy Hacking! 💯 💯 💯

src/main/kotlin/blackjack/model/Suit.kt
miooo0o: ah oh, yeah. I got it. somehow make sense.
I will reflecting this in step 2.
Thank you!

src/test/kotlin/blackjack/model/CardDeckTest.kt
miooo0o: `However, that doesn't mean the CardDeck is free from validating it's own size before a hit.`

Very clear answer. I would like to apply this approach in step 2... 
Thank you.

src/main/kotlin/blackjack/controller/GameManager.kt
miooo0o: Oh really? Good thing I found out before step2. Thank you so much, you save me.

I was actually thinking about breaking up `GameManager` even more.

while waiting for the review, I reverted and reworked it for my practice, and it's easier to read and more concise... 😭 For step2, I'll try to keep it as simple as possible and focus on meeting the minimum requirements.

PR #16, opened 2025-07-14
[Blackjack - step1] Submit knopgm's mission

knopgm:
# Blackjack Project -Step1

## Overview:
This project aim to mimic the basic rules of a Blackjack game.
It's following the rules of OOP and a approach towards RDD.

## Main Features:
- Classes: [Participant, Player, Dealer, Card, Deck, Hand]
- Views: [ InputView, OutputView, InputValidator ]
- Controllers: [ BlackJack, GameLogic]
- Utils: [ CardSuit, ResultTypes ]

`BlackJack` is the main connection with all functionalities.

`Participant` is an abstract function that will extends as `Player` and `Dealer`.

`Card` defines what one card is and Deck define what list of cards should be created to be valid as a game deck.

`Hand` is the one who knows what cards a hand has and its score.

`GameLogic` was used to build the steps the game needs.

### Current Limitations:
I still have to do Tests for many Functionalities. (Schedule-time)
I know my conditional functions didn't use the rules of Object Calisthenics, and I'll try to improve it in the review code time. (Schedule-time)
I loved to understand and use the abstract class. 
While I was coding huge interactions with a list of players or participants or their related list outputs, I had a tip to interact with only one item of the list at a time, which reduce a lot the use of conditionals and the size of my methods. But I found it hard to decide which one of the involved classes were responsible for some behaviors. -> Like for example in the method `checkWinOrLose`: Who is responsible to check it? The dealer, the participant or the hand? 
As a beginner, It still blurry how to build a code that is clear for any other developer to read and follow. I think architecture will make a difference to a easier understand. Does it? and how much it do make a difference?
_________________
Thank you for your time and thoughtful guidance in advance! 🙏


README.md
```diff
@@ -1 +1,34 @@
...
+
+### Dealer
+- [x] Draw card to any participant
+- [x] Dealer auto-draws when score ≤ 16
+- [x] Ask players if they want to draw more cards
+
+### Hand
+- [x] Calculate participant score with Ace logic
+- [x] Calculate participant score with J, Q and K logic
+- [ ] provide information for game results announce
```
Wordbe, line 25: Please check this if you comleted!

README.md
```diff
@@ -1 +1,34 @@
...
+
+### Game
+- [ ] Accept player names
```
Wordbe, line 28: Same here!

src/main/kotlin/blackjack/BlackjackController.kt
```diff
@@ -0,0 +1,31 @@
...
+        players.forEach { player ->
+            output.printFirstTurnCards(player)
+        }
+
+        game.otherTurnCards(players, dealer)
+
+        output.printFinalScores(dealer)
+        players.forEach { player -> output.printFinalScores(player) }
+
+        println("## Final Results ##")
```
Wordbe: Displaying output is the responsibility of the View, so let the View handle it! 🙂

src/main/kotlin/blackjack/InputView.kt
```diff
@@ -0,0 +1,19 @@
+package blackjack
+
+class InputView {
+    private val validator = InputValidator()
+    val output = OutputView()
```
Wordbe: Try to use OutputView only within the controller.
It's fine for InputView to use println to guide user input! 🙂

src/main/kotlin/blackjack/GameLogic.kt
```diff
@@ -0,0 +1,35 @@
+package blackjack
+
+class GameLogic {
+    fun namesToPlayers(names: List<String>): List<Player> {
```
Wordbe, line 4: You can simplify this into a one-liner by removing the variable!

src/main/kotlin/blackjack/GameLogic.kt
```diff
@@ -0,0 +1,35 @@
...
+            }
+        })
+        repeat(2, {
+            val card = dealer.drawCard()
+            dealer.addCard(card)
+        })
+    }
+
+    fun otherTurnCards(players: List<Player>, dealer: Dealer) {
+        val output = OutputView()
```
Wordbe: How about making the interaction between the View and the domain model occur only in the Controller? 🤔

src/main/kotlin/blackjack/Hand.kt
```diff
@@ -0,0 +1,34 @@
...
+    val cards: MutableList<Card> = mutableListOf()
+
+    fun addCard(card: Card) {
+        cards.add(card)
+    }
+
+   fun sumCards(): Int {
+        val cardValues = cards.map { card -> card.number }.toMutableList()
+        val elevenSumCheck = cardValues.sum() + 10
+        cardValues.replaceAll { if (it == 1 && elevenSumCheck <= 21) 11 else it }
```
Wordbe: It would be good to define magic numbers as constants! 
Aside from what's inside this class, please change everything else that’s visible!

src/main/kotlin/blackjack/Hand.kt
```diff
@@ -0,0 +1,34 @@
...
+        val cardValues = cards.map { card -> card.number }
+        val score = cardValues.sum()
+        return when {
+            score == 21 -> "Win"
+            score > 21 -> "Lose"
+            else -> "Stay"
+        }
+    }
+
+    override fun toString(): String {
```
Wordbe: If you create elements needed for the view inside a domain object, it introduces a dependency on the view. This means that when the view changes, the domain object must also be modified. Try moving this logic into the view instead! 🙂

src/main/kotlin/blackjack/Dealer.kt
```diff
@@ -0,0 +1,18 @@
...
+class Dealer(val deck: Deck) : Participant("Dealer") {
+    fun drawCard(): Card {
+        return deck.drawCard()
+    }
+
+    override fun shouldHit(): Boolean {
+        val scoreHand = this.sumCards()
+        if (scoreHand < 17) {
+            println()
+            println("Dealer draws one more card due to having 16 or less.")
```
Wordbe: Instead of having OutputView handle this part, try throwing an exception here. 😊

src/main/kotlin/blackjack/Deck.kt
```diff
@@ -0,0 +1,24 @@
+package blackjack
+
+class Deck(cards: List<Card>) {
+    val cards = cards.shuffled().toMutableList()
```
Wordbe: You can declare properties together in the primary constructor. Try combining them.

Also, functions like shuffled() that return random results make testing difficult. How about moving that logic outside of Deck so it’s easier to test? 😊

src/main/kotlin/blackjack/Hand.kt
```diff
@@ -0,0 +1,34 @@
+package blackjack
+
+class Hand() {
+    val cards: MutableList<Card> = mutableListOf()
```
Wordbe: Try creating it in the primary constructor—it improves readability and makes testing easier.

src/main/kotlin/blackjack/InputValidator.kt
```diff
@@ -0,0 +1,10 @@
+package blackjack
+
+class InputValidator {
+    fun validateNames (input: List<String>) {
+        input.forEach { name ->
+            require(name.isNotBlank()) { "Input cannot be blank." }
+            require(name.chars().count() > 2) { "Name should have more than 2 characters." }
```
Wordbe: There should be a function to get the length of a string. Try using it!

src/main/kotlin/blackjack/Participant.kt
```diff
@@ -0,0 +1,27 @@
...
+        return when {
+            result == 21 -> ResultTypes.BLACKJACK
+            result > 21 -> ResultTypes.BUSTED
+            else -> ResultTypes.STAY
+        }
+    }
+
+    abstract fun shouldHit(): Boolean
+
+
```
Wordbe: Please remove the blank spaces!

src/main/kotlin/blackjack/Player.kt
```diff
@@ -0,0 +1,16 @@
+package blackjack
+
+class Player(playerName: String): Participant(playerName) {
+    override fun shouldHit(): Boolean {
+        val scoreHand = this.sumCards()
+        if (scoreHand < 21) {
+            println()
+            println("Would ${this.name} like to draw another card? (y for yes, n for no)")
+            val wantHit = readln().trim()
```
Wordbe: The InputView is performing tasks that should be handled within the domain object. It looks like a change is needed. 
(println, readln)

src/main/kotlin/blackjack/ResultTypes.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack
+
+enum class ResultTypes() {
```
Wordbe: You can omit the default constructor since it’s automatically generated. 🙂

src/test/kotlin/blackjack/DealerTest.kt
```diff
@@ -0,0 +1,15 @@
...
+import org.junit.jupiter.api.Test
+
+class DealerTest {
+    @Test
+    fun `draws a card`() {
+        val deck = Deck.generate()
+        Dealer(deck).drawCard()
+        val deckSize = deck.cards.size
+
+        assertEquals(47, deckSize, "Should draw a card out of the deck")
```
Wordbe, line 30: Don't just test the count—create specific cards and check if they match as expected.
If you extract the random generation part, it becomes testable.

Wordbe (changes requested):
Hello Gabi!

Thank you for diligently working on the Blackjack mission and submitting your PR. 🙌

I've left comments on areas that might be worth reconsidering. Please take a look and reflect any changes you find appropriate!

Also, the tests could use some reinforcement—I'd appreciate it if you could enhance them.

Thanks again! 😊

src/main/kotlin/blackjack/GameLogic.kt
knopgm: I can see that my code was violation the responsibility rules of the logic, so I refactor a lot of things about the input/output view (again) in the same commit.
I hope now all the responsibilities are more clear 😅

src/main/kotlin/blackjack/Deck.kt
knopgm: I created a boolean parameter for the deck generator function, so I can use the function with false to test it, and the function with true to the game logic 😅  I hope that fits the solution you are looking for. 🙏

src/test/kotlin/blackjack/DealerTest.kt
knopgm: I created a test with specific cards to see if the right card is draw. :)
I'll try to implement more test later.

knopgm:
Hy [Wordbe](https://github.com/Wordbe)!
Thank you for your time and detailed review! I really felt my learn improving by seeing what was missing/ wrong while I solved all your comments. 🙏 

The main refactor of my review was separation of concerns, which I regret not to have cared before (time constrains).
Nevertheless it took me some time to move everything to their places, so I'm glad that I had the time to dive into the architecture of blackjack and figure out the concerns with time.

Let me know if any additional improvements are needed! 😄

src/main/kotlin/blackjack/GameLogic.kt
```diff
@@ -0,0 +1,32 @@
+package blackjack
+
+class GameLogic {
+    fun namesToPlayers(names: List<String>): List<Player> {
+         return names.map { name -> Player(name) }
```
Wordbe, line 5: Great 👍

src/main/kotlin/blackjack/GameLogic.kt
Wordbe: Great. Thank you!

src/main/kotlin/blackjack/Deck.kt
Wordbe: What I meant wasn’t to add a new field, but rather to move out behavior like `shuffled()`—which has an uncontrollable return value—from the object and pass it in as a parameter instead. That way, your code will be much easier to test! 😊

src/main/kotlin/blackjack/Card.kt
```diff
@@ -0,0 +1,19 @@
...
+            11 -> "Q"
+            12 -> "K"
+            else -> number.toString()
+        }
+    }
+
+    override fun toString(): String {
+        val cardValue = numberToCardDeckElements(number)
+        return "%s%s".format(cardValue, suit.value)
+    }
```
Wordbe, line 17: The domain object is currently handling code that's only needed in the View layer. Try to decouple the domain from the View by moving that logic into the View layer instead. 🙂

Wordbe (approved):
Hello!

Great job handling the reviews for the Blackjack mission.
It was nice to see how you abstracted `Participant` and extended it into `Player` and `Dealer`.

Please revisit the feedback I provided and try to apply it throughout the project as you go.
Make your code even more concise with Kotlin, aim for a clear separation between Model and View, and reinforce your tests overall.
Also, take a look at today’s review comments and reflect on what can be improved.

See you in step2. Great work! 👍

PR #17, opened 2025-07-14
[blackjack - Step2] Submit Apoorva's mission

Apoorva-S-Natesh:
Hello,
Thank you for taking time to review this code.
This PR is regarding the extended part of blackjack game where each player can bet and will earn or lose money according to the game results.

The bet amount is wrapped in Bet
player score > 21 -> lose entire bet
    - player score = 21 (BlackJack) -> 1.5x their bet
    - dealer score = 21 (blackJack), player score = 21 (not blackJack) -> lose entire bet
    - player score = 21 (not blackJack) -> player’s bet is returned
    - player == dealer and < 21 -> player’s bet is returned.
    - dealer > 21 (bust) -> player’s bet is returned.
    
Questions:
Is there a better way to write the tests for calculating earnings?

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,12 @@
 
 ## betting
 
-- [ ] prompt each user for betting amount
+- [x] prompt each user for betting amount
     - amount > 0
-- [ ] Calculate final earnings after the game
+- [x] Calculate final earnings after the game
     - player score > 21 -> lose entire bet
     - player score = 21 (BlackJack) -> 1.5x their bet
     - player score = 21 (not blackJack) -> player’s bet is returned
     - player == dealer and < 21 -> player’s bet is returned.
     - dealer > 21 (bust) -> player’s bet is returned.
-- [ ] Display final earnings
+- [x] Display final earnings
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@
 - [x] Calculate final earnings after the game
     - player score > 21 -> lose entire bet
     - player score = 21 (BlackJack) -> 1.5x their bet
+    - dealer score = 21 (blackJack), player score = 21 (not blackJack) -> lose entire bet
     - player score = 21 (not blackJack) -> player’s bet is returned
     - player == dealer and < 21 -> player’s bet is returned.
     - dealer > 21 (bust) -> player’s bet is returned.
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 - [x] prompt user for players names
 - [x] players name parsing
 - [x] Prompt for user decision (y/n)
-- [x] decision parsing 
+- [x] decision parsing
 
 ## Player
 
@@ -44,4 +44,16 @@
 - [x] display dealers draw action
 - [x] display players done action
 - [x] display all players hands with total score
-- [x] display final result 
+- [x] display final result
+
+## betting
+
+- [ ] prompt each user for betting amount
+    - amount > 0
+- [ ] Calculate final earnings after the game
+    - player score > 21 -> lose entire bet
+    - player score = 21 (BlackJack) -> 1.5x their bet
+    - player score = 21 (not blackJack) -> player’s bet is returned
+    - player == dealer and < 21 -> player’s bet is returned.
+    - dealer > 21 (bust) -> player’s bet is returned.
+- [ ] Display final earnings
--- a/src/main/kotlin/controller/BlackJackController.kt
+++ b/src/main/kotlin/controller/BlackJackController.kt
@@ -1,5 +1,6 @@
 package controller
 
+import model.Bet
 import model.Dealer
 import model.Player
 import model.Players
@@ -13,6 +14,7 @@ object BlackJackController {
         try {
             val players = Players(getPlayerNames().map { Player(it) })
             val dealer = Dealer()
+            startBetting(players)
             startGame(players, dealer)
             runTurns(players, dealer)
             showResults(players, dealer)
@@ -26,6 +28,11 @@ object BlackJackController {
         val players = InputView.requestPlayerNames()
         return players
     }
+
+    private fun startBetting(allPlayers: Players) {
+        allPlayers.players.forEach { player -> player.bet = Bet(InputView.requestPlayersBet(player.name)) }
+    }
+
 
     private fun startGame(
         players: Players,
@@ -62,5 +69,8 @@ object BlackJackController {
             dealer.getScore(),
         )
         OutputView.displayResults(result, allPlayers)
+        val earningResult = ResultCalculator.calculateEarning(allPlayers, dealer.getScore(), dealer.isBlackJack())
+        ResultCalculator.applyEarningResult(allPlayers, dealer, earningResult)
+        OutputView.displayFinalEarnings(allPlayers, dealer)
     }
 }
--- a/src/main/kotlin/controller/BlackJackController.kt
+++ b/src/main/kotlin/controller/BlackJackController.kt
@@ -30,7 +30,6 @@ object BlackJackController {
     }
 
     private fun startBetting(allPlayers: Players) {
-
         allPlayers.players.forEach { player ->
             val betInput = InputView.requestPlayersBet(player.name)
             player.bet = Bet.parseToBet(betInput)
--- a/src/main/kotlin/controller/BlackJackController.kt
+++ b/src/main/kotlin/controller/BlackJackController.kt
@@ -19,7 +19,7 @@ object BlackJackController {
             runTurns(players, dealer)
             showResults(players, dealer)
         } catch (e: IllegalArgumentException) {
-            println(e.message)
+            OutputView.displayErrorMessage(e)
             playGame()
         }
     }
@@ -50,7 +50,7 @@ object BlackJackController {
         dealer: Dealer,
     ) {
         allPlayers.players.forEach { player ->
-            while (InputView.requestPlayerDecision(player.name).isYes()) {
+            while (player.getScore() <= 21 && InputView.requestPlayerDecision(player.name).isYes()) {
                 player.requestCardFromDealer(dealer)
                 OutputView.displayPlayersTurn(player)
             }
--- a/src/main/kotlin/controller/BlackJackController.kt
+++ b/src/main/kotlin/controller/BlackJackController.kt
@@ -31,8 +31,7 @@ object BlackJackController {
 
     private fun startBetting(allPlayers: Players) {
         allPlayers.players.forEach { player ->
-            val betInput = InputView.requestPlayersBet(player.name)
-            player.bet = Bet.parseToBet(betInput)
+            player.bet = Bet(InputView.requestPlayersBet(player.name))
         }
     }
 
--- a/src/main/kotlin/controller/BlackJackController.kt
+++ b/src/main/kotlin/controller/BlackJackController.kt
@@ -33,7 +33,6 @@ object BlackJackController {
         allPlayers.players.forEach { player -> player.bet = Bet(InputView.requestPlayersBet(player.name)) }
     }
 
-
     private fun startGame(
         players: Players,
         dealer: Dealer,
@@ -64,10 +63,11 @@ object BlackJackController {
         dealer: Dealer,
     ) {
         OutputView.displayFinalCardsOnHand(allPlayers, dealer)
-        val result = ResultCalculator.getResult(
-            allPlayers.players.map { it.getScore() },
-            dealer.getScore(),
-        )
+        val result =
+            ResultCalculator.getResult(
+                allPlayers.players.map { it.getScore() },
+                dealer.getScore(),
+            )
         OutputView.displayResults(result, allPlayers)
         val earningResult = ResultCalculator.calculateEarning(allPlayers, dealer.getScore(), dealer.isBlackJack())
         ResultCalculator.applyEarningResult(allPlayers, dealer, earningResult)
--- a/src/main/kotlin/controller/BlackJackController.kt
+++ b/src/main/kotlin/controller/BlackJackController.kt
@@ -2,6 +2,7 @@ package controller
 
 import model.Dealer
 import model.Player
+import model.Players
 import service.ResultCalculator
 import util.isYes
 import view.InputView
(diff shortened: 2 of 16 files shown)


src/main/kotlin/service/ResultCalculator.kt
```diff
@@ -13,6 +15,40 @@ class ResultCalculator {
...
+                    when {
+                        dealerScore > 21 -> bet to -bet
+                        dealerHasBlackJack && !playerHasBlackJack -> -bet to bet
+                        playerHasBlackJack && dealerHasBlackJack -> 0.0 to 0.0
+                        playerScore == dealerScore -> 0.0 to 0.0
+                        playerHasBlackJack -> bet * 1.5 to (-bet * 1.5)
+                        playerScore > dealerScore && playerScore <= 21 -> bet to -bet
+                        else -> -bet to bet
+                    }
+                PlayerEarningResult(index, playerEarning, dealerEarning)
```
Apoorva-S-Natesh, line 37: when I was trying to use map index, I had a question about Why does map.indexed return `List<Any>` instead of `List<MyObject>` but the error was resolved when an else was added to when block?

src/main/kotlin/service/ResultCalculator.kt
```diff
@@ -13,6 +15,40 @@ class ResultCalculator {
...
+        }
+
+        fun applyEarningResult(
+            allPlayers: Players,
+            dealer: Dealer,
+            earnings: List<PlayerEarningResult>,
+        ) {
+            earnings.forEach { earning ->
+                allPlayers.players[earning.playerId].earnings += earning.earningsChange
+                dealer.earnings += earning.dealerEarningChange
```
Apoorva-S-Natesh: How many levels of reference can I go?
is there a better way to achieve this without so much nesting?

README.md
```diff
@@ -44,4 +44,17 @@
...
+- [x] prompt each user for betting amount
+    - amount > 0
+- [x] Calculate final earnings after the game
+    - player score > 21 -> lose entire bet
+    - player score = 21 (BlackJack) -> 1.5x their bet
+    - dealer score = 21 (blackJack), player score = 21 (not blackJack) -> lose entire bet
+    - player score = 21 (not blackJack) -> player’s bet is returned
+    - player == dealer and < 21 -> player’s bet is returned.
+    - dealer > 21 (bust) -> player’s bet is returned.
+- [x] Display final earnings
```
Hyeon9mak, line 60: 😉 👍

src/main/kotlin/controller/BlackJackController.kt
```diff
@@ -1,29 +1,43 @@
...
-        showResults(players, dealer)
+        try {
+            val players = Players(getPlayerNames().map { Player(it) })
+            val dealer = Dealer()
+            startBetting(players)
+            startGame(players, dealer)
+            runTurns(players, dealer)
+            showResults(players, dealer)
+        } catch (e: IllegalArgumentException) {
+            println(e.message)
```
Hyeon9mak: 😱 It seems like view's responsibility.

src/main/kotlin/controller/BlackJackController.kt
```diff
@@ -1,29 +1,43 @@
...
-        startGame(players, dealer)
-        runTurns(players, dealer)
-        showResults(players, dealer)
+        try {
+            val players = Players(getPlayerNames().map { Player(it) })
+            val dealer = Dealer()
+            startBetting(players)
+            startGame(players, dealer)
+            runTurns(players, dealer)
+            showResults(players, dealer)
```
Hyeon9mak, line 20: I really like how you've organized the game flow in a clear sequence.
It makes the code very readable! 💯

src/main/kotlin/model/BasePlayer.kt
```diff
@@ -1,19 +1,16 @@
...
 
-    fun getCardsNumber(): Int = hand.getCards().size
-
     fun showCards(): Set<Card> = hand.getCards()
 
     fun drawCard(card: Card) {
         hand.addCard(card)
     }
+
+    fun isBlackJack(): Boolean = getScore() == 21 && showCards().size == 2
```
Hyeon9mak, line 15: 💯 👍 now it can check REAL BlackJack!

src/main/kotlin/model/Bet.kt
```diff
@@ -0,0 +1,11 @@
...
+
+data class Bet(val amount: Double) {
+    companion object {
+        fun parseToBet(input: String): Bet {
+            val amount = input.trim().toDoubleOrNull() ?: throw IllegalArgumentException("Bet must be a valid number")
+            if (amount < 0) throw IllegalArgumentException("Amount should be greater than 0")
+            return Bet(amount)
+        }
+    }
+}
```
Hyeon9mak: I think it can change to value class 😉 
and how about move validation logics to constructor?
then it can get confidence with used constructor.

And, I wonder if handling string input or null values might be more appropriate in the view layer rather than in the model.

```suggestion
data class Bet(val amount: Double) {

    init {
        require(amount >= 0) { "Amount should be greater than 0" }
    }
    
    companion object {
        // ...
    }
}
```

src/main/kotlin/view/OutputView.kt
```diff
@@ -23,23 +24,34 @@ object OutputView {
...
+
+    fun displayFinalEarnings(
+        allPlayers: Players,
+        dealer: Dealer,
+    ) {
+        println("\n## Final Earnings")
+        println("Dealer: ${dealer.earnings.toInt()}")
+        for (i in allPlayers.players.indices) {
+            println("${allPlayers.players[i].name}: ${(allPlayers.players[i].earnings).toInt()}")
         }
```
Hyeon9mak, line 55: Finally! Now we can focus on the game without even knowing the dealer’s name!
And if we ever want to change how the dealer appears, we only need to look at the view 👍

src/main/kotlin/model/Dealer.kt
```diff
@@ -1,11 +1,11 @@
 package model
 
-class Dealer() : BasePlayer("Deal") {
+class Dealer() : BasePlayer() {
```
Hyeon9mak:
```suggestion
class Dealer : BasePlayer() {
```
😉

src/main/kotlin/service/ResultCalculator.kt
Hyeon9mak: > If your when expression doesn't have a subject, you must have an else branch or the compiler throws an error. The else branch is evaluated when none of the other branch conditions are satisfied
>
> https://kotlinlang.org/docs/control-flow.html#when-expressions-and-statements

That is reason of need to make it a habit to read the official docs 😄

src/main/kotlin/service/ResultCalculator.kt
Hyeon9mak: If not too much bothering, it’s a good idea to create a separate method to keep the reference level as low as possible.

The fewer getters you use (and the fewer classes that become interconnected),
the easier the code is to maintain.

That’s what we call **encapsulation**.

```kotlin
allPlayers.appendEarnings(earning.earningsChange)

// Players.kt
class Players(val players: List<Player>) {
    init {
        require(players.size <= 6) { "Maximum player names must be 6" }
    }

    fun appendEarnings(double: Double) {
        players.forEach { it.earnings += double }
    }
}
```

Hyeon9mak (changes requested):
Hi Apoorva! Great work in step2 👍 

The betting requirement wasn’t too tricky, right?
I left quite a few comments to help improve the overall code quality.

Also, it looks like there's a small gap in the game rules — which gave rise to a greedy Apoorva 😄
It would be great if you could patch that up too!

image

Plz feel free to leave comments 😉

src/main/kotlin/service/ResultCalculator.kt
Apoorva-S-Natesh: Thank you for the explanation! Should Have read the official docs.
If I understood it right
Without else: Kotlin can't infer a consistent return type → List<Any>
With else: Kotlin sees a consistent return type → List<MyObject>

src/main/kotlin/controller/BlackJackController.kt
Apoorva-S-Natesh: Got it! Any display or printing should be handled by the view. Thank you for catching it

src/main/kotlin/controller/BlackJackController.kt
Apoorva-S-Natesh: Thank you!

src/main/kotlin/view/OutputView.kt
Apoorva-S-Natesh: Starting with eCommerce mission, I understand your point about not giving Dealer a name and letting the view decide it.

src/main/kotlin/service/ResultCalculator.kt
Apoorva-S-Natesh: Thanks for the suggestion — I agree that reducing reference depth and improving encapsulation makes the code more maintainable.

I initially considered a method like `appendEarnings(Double)` as you proposed, but in this case, each player receives a different earningsChange, so applying the same value to all players wouldn't reflect the actual logic.

To keep the reference level low while preserving the per-player earnings logic, I added a method to the `Players class` that takes a list of `PlayerEarningResult` and applies the changes internally. This method also updates the dealer’s earnings as part of the same loop, since the dealer’s earnings are tied to each player's result.

This way, the logic is still encapsulated, and the external call remains clean: `allPlayers.appendEarnings(results, dealer)`.

src/main/kotlin/model/Bet.kt
Apoorva-S-Natesh: You're absolutely right — since Bet is a value class, it makes sense to lean into its strengths and ensure that all instances are valid by design. Moving the validation logic directly into the constructor (via init) would indeed make the class more self-contained and trustworthy, especially when used across different parts of the codebase.

That said, Kotlin's value class constructors can't be made private, so while we can enforce validation in the init block, we can't fully prevent misuse unless we wrap creation in a factory method. It's a trade-off between convenience and safety. What would you suggest?

As for handling string inputs and nulls — I completely agree that parsing and input validation (like checking for empty strings or malformed numbers) is better suited for the view layer.

src/main/kotlin/service/ResultCalculator.kt
Hyeon9mak: Kotlin allows you to omit else only when it has nothing left to consider.
But if there's still a possibility Kotlin can't rule out, else becomes mandatory.

For example:

- When using an enum class Suit and only handling HEART and CLUB, you must include else.
- When using all enum values (HEART, CLUB, DIAMOND, SPADE), Kotlin knows there's nothing left, so else can be omitted.

src/main/kotlin/model/Players.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    fun appendEarnings(
+        results: List<PlayerEarningResult>,
+        dealer: Dealer,
+    ) {
+        results.forEach { earning ->
+            players[earning.playerId].earnings += earning.earningsChange
+            dealer.earnings += earning.dealerEarningChange
+        }
+    }
```
Hyeon9mak, line 18: 😉 👍 
now when we need change for append earning logic,
we just check this class!

src/main/kotlin/model/Bet.kt
```diff
@@ -0,0 +1,15 @@
...
+        require(amount >= 0) { "Amount should be greater than 0" }
+    }
+
+    companion object {
+        fun parseToBet(input: String): Bet {
+            val amount = input.trim().toDouble()
+            return Bet(amount)
+        }
+    }
+}
```
Hyeon9mak: > Kotlin's value class constructors can't be made private, so while we can enforce validation in the init block, we can't fully prevent misuse unless we wrap creation in a factory method. It's a trade-off between convenience and safety. What would you suggest?

I think the View layer can handle null or blank inputs and convert them to `Double`.
Just like -- how browsers send a `Boolean` instead of a string when toggling switch button.

Hyeon9mak (changes requested):
Hi Apoorva!

All the requirements have been met, and the program is working well!
I just left very minor comments — feel free to take a look and try it out at your own pace.

I heard Jason recently introduced the State pattern —
I figured you might have some questions about it or be interested in trying it out,
so I left final 'Request Changes'. 😉 

If you don't have any major questions, feel free to request a review again right away!

src/main/kotlin/model/Bet.kt
Apoorva-S-Natesh: wow, that's a good example! Had never thought about that. I always thought that the View should always be responsible for just taking input and the parsing should not be done by the View. But this makes a lot of sense now! thank you

src/main/kotlin/model/Players.kt
Apoorva-S-Natesh: Thank you for helping me understand the concept of separation of concerns. Your explanation about how a teammate working on a feature only needs to focus on a specific part of the code really clarified things for me.

Apoorva-S-Natesh: Kotlin’s requirement for else depending on whether all possibilities are covered is a smart way to ensure exhaustive handling. Using enums like that really helps clarify how Kotlin enforces safety. Appreciate the example!

Apoorva-S-Natesh: Thanks for the feedback and for pointing out the State pattern suggestion! At this point, I’d prefer not to refactor the code to use the State pattern just yet, as I’m still getting familiar with it and would like to take some time to understand it better before applying it.  I’m definitely going to explore it in the future and refactoring when I feel more confident with the pattern. I have a lot of confusion and doubts that I would like to explore to understand the design pattern.
1. This might be very silly but - How is the State pattern different from using simple if-else or when statements?
2. How do I decide what should be a "state" in my application?
3. Where should the logic for transitioning between states live 
4. How does the State pattern help with scalability or maintainability in the long run?

 Appreciate your guidance and the opportunity to learn!

Hyeon9mak: 1. In the end, if you think of it simply as a file system, it's just a set of class files that hide internal logic using if-else statements.
2. You can represent the game progress status of players and dealers — for example: init, hit, bust, blackjack, etc.
3. Design patterns exist to improve maintainability. That's why "state" should also be viewed from a business perspective, not just a technical one. Since the states belong to the player and dealer, it makes sense to place them inside the Participant.
4. The State Pattern is especially helpful when each state contains complex logic (e.g., more than 10 lines inside an if block) or when a specific state is frequently modified. For example, if a teammate needs to update the bust rule, they can just go straight to the Bust class.

Hyeon9mak (approved):
Hi Apoorva!
Great job reflecting all the feedback!

Since all requirements are fully met, I’m approving this right away.

Hope you enjoy the rest of the missions too! 😉

PR #18, opened 2025-07-14
[blackjack - Step 2] Submit mehrshahbaz's mission

MehrShahbaz:
- I would like get a review on how can I make this code so it looks like it is from a mid level kotlin developer, any changes or repetitions which could be removed by using Kotlin magic


src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -1,20 +1,26 @@
 package blackjack.controller
 
-import blackjack.model.*
+import blackjack.model.Card
+import blackjack.model.Dealer
+import blackjack.model.Deck
+import blackjack.model.Gambler
+import blackjack.model.GamblerInfo
+import blackjack.model.Player
```
pkeugine, line 8: Nice job [applying previous review about wildcard](https://github.com/herotechcourse/kotlin-blackjack/pull/18/commits/08a3c8108c21816ba65a90fc623c0569ecb80dc2) 👍 

Why do you think I suggested this?
It's great to know the reason :)

Also, the commit I've linked contains works that are not related to wildcard.
It'll be great if we could separate works and add them to a more appropriate commit.
You don't have to do it now, just a reminder so we can think about it in the next mission.

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -1,16 +1,29 @@
...
         get() = _cards
+    val winnings: Double
+        get() = playerBet.winnings
 
     val name: String
         get() = gamblerInfo.name
 
+    fun setBetAmount(amount: Double) {
+        playerBet.betAmount = amount
+    }
```
pkeugine: Instead of making a class with `var` field and a `setXXX`,
why don't we inject it via controller as the object gets created?
I'm thinking something like `Player(gamblerInfo, playerBet)`.
This way, we'll be able to manage them immutably.

This would also mean that `Dealer`'s `calculateAndSetWinnings` would also get changed.

src/test/kotlin/blackjack/DeckTest.kt
```diff
@@ -32,4 +32,13 @@ class DeckTest {
...
     }
+
+    @Test
+    fun `Generated new cards if one card left`() {
+        val deck = Deck()
+        deck.drawCards(51)
+        deck.drawCards()
+
+        assertThat(deck.cards.size).isEqualTo(51)
+    }
```
pkeugine: This is just a question about the way game works:

If there's only one card left, does that card get discarded, and new set of deck get created?

src/test/kotlin/blackjack/GamblerTest.kt
```diff
@@ -20,4 +25,206 @@ class GamblerTest {
...
         assertThat(gambler.isPlayerBelowBlackJack()).isFalse
     }
+
+    @Test
+    fun `return true if Player has two cards`() {
+        val gambler = Gambler(GamblerInfo("Player"))
+        val cards = listOf(Card(Rank.ACE, Suit.SPADE), Card(Rank.KING, Suit.SPADE))
+        gambler.addCard(cards)
+        assertThat(gambler.hasCardCount()).isTrue
+    }
```
pkeugine: It would be great to add 'what returns true' in test name. In this case, I guess it would be `hasCardCount`.

Personally, I like to keep method names into test titles since it is easier to search for it, and it allows the team to automate various processes when managing a project.

src/test/kotlin/blackjack/GamblerTest.kt
```diff
@@ -20,4 +25,206 @@ class GamblerTest {
...
+        )
+
+        val player = Gambler(GamblerInfo("Player"))
+        player.setBetAmount(10.0)
+        player.addCard(
+            listOf(Card(Rank.ACE, Suit.SPADE), Card(Rank.SIX, Suit.SPADE)),
+        )
+        player.setWinnings(dealer)
+        assertThat(player.winnings).isEqualTo(-10.0)
+    }
```
pkeugine: Wow, amazing work adding in test cases 💯 💯

README.md
```diff
@@ -5,7 +5,6 @@
 ### Data class `Card`
```
pkeugine, line 5: **Q. I would like get a review on how can I make this code so it looks like it is from a mid level kotlin developer, any changes or repetitions which could be removed by using Kotlin magic**

You're actually doing great, and I see that your project doesn't have much to comment on.
It's clean and looks easy to maintain.

I wouldn't focus too much on trying to become a `mid level Kotlin developer`, and to be honest, I think we need to study **at least** 2~3 more years in order to become one. There's so many things that we need to cover, and a lot of practice so that we can apply them to our everyday programming.

Anyways, your code looks great, so don't worry about it. 👍

pkeugine (changes requested):
Hello Shahbaz 🙂 
Nice work on step 2!
It was really hard to find spots where I could comment on. 👍 

I left some, so go over them when you have the time.
Ask me anything via comments or DM when you have one!
And like always,
Happy Hacking 🔥 🔥 🔥

src/main/kotlin/blackjack/controller/Controller.kt
MehrShahbaz: You suggested this because explicit imports are better for readability and maintainability

pkeugine: That's right.
When using wildcard, we import unused logic as well which is unnecessary.
Also, since wildcard related options are different from a developer to another,
there's a high chance for us to meet a conflict when working with teammates. :)

pkeugine (approved):
Hello Shabaz!
Nice work on additional refactoring. 👍 
I think it's time we finish blackjack and move on to the next mission.

It was fun reviewing your project, and I hope you do as good as this mission on your next one. Keep up the good work!

Happy Hacking! 🚄 🚄 🚄

PR #19, opened 2025-07-14
[blackjack - Step2] Submit SebasNadu's mission

SebasNadu:
Hi Wordbe! 👋

This PR adds betting mechanics and improves class design in the Blackjack game. I also made several refactorings to improve readability, encapsulation, and separation of concerns.

I didn’t go forward with the state pattern this time, I’ve parked it on my TODO list so I can continue with the course flow. That said, I’d love to hear your thoughts and learn more from your perspective as always!

### Betting Mechanics:
* Introduced a new `Chips` value class to manage betting amounts, ensuring immutability and providing operations like addition, subtraction, and multiplication for bet calculations (`src/main/kotlin/blackjack/model/Chips.kt`).
* Added a `placeBets` step in the game flow to allow players to place bets before the game starts.
* Updated the `Result` enum to include a `BLACKJACK` result for handling blackjack-specific outcomes.

### Class Design Improvements:
* Updated the `Card` class to implement `Comparable`, enabling sorting of cards by rank, after the first two cards.
* Added a `sortCards` method in the `Hand` class and a corresponding `sortHand` method in the `Participant` class to sort cards for better readability (`src/main/kotlin/blackjack/model/card/Hand.kt`.

### Codebase Simplification:
* Removed redundant `toString()` implementations in several classes, delegating responsibility for string representation to the view layer.
* Changed internal methods in `BlackJackGame` to private to encapsulate functionality and improve code safety, when they are not using for testing reasons.

### View and Input Enhancements:
* Updated `InputView` to include a method for retrieving betting amounts and validating that the input is a positive number.
* Adjusted `OutputView` to display final earnings instead of results, showing profits for both the dealer and players.

### TODO
* Implement the state pattern in Participant and centralize the result and profit logic there.

### Example

Screenshot 2025-07-14 at 15 20 01


README.md
```diff
@@ -58,11 +63,18 @@
     - [x] Should be comma separated names
     - [x] Should contain a minimum of one player.
     - [x] The names should be unique.
+  - [ ] Should accept an amount for the Bet bigger than 0.
```
Wordbe: Check this please!

src/main/kotlin/blackjack/model/Chips.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    fun blackjack(): Chips = this * BLACKJACK_MULTIPLIER
+
+    operator fun unaryMinus(): Chips = Chips(-value)
+
+    operator fun times(multiplier: Double): Chips = Chips(value * multiplier)
+
+    operator fun plus(other: Chips): Chips = Chips(value + other.value)
+
+    override fun toString(): String = "${value.toInt()}"
```
Wordbe: If it's being used in the view, it would be better to place it within the view. 👀

src/main/kotlin/blackjack/model/Chips.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    companion object {
+        private const val BLACKJACK_MULTIPLIER = 1.5
+        private const val ZERO = 0.0
+
+        fun zero(): Chips = Chips(ZERO)
```
Wordbe: How about creating Chips.ZERO?

src/main/kotlin/blackjack/model/Chips.kt
```diff
@@ -0,0 +1,23 @@
+package blackjack.model
+
+@JvmInline
+value class Chips(val value: Double) {
```
Wordbe, line 4: Good idea, value class!

src/main/kotlin/blackjack/model/GameJudge.kt
```diff
@@ -10,12 +10,14 @@ object GameJudge {
...
-    internal fun evaluate(
+    private fun evaluate(
         dealer: Dealer,
         player: Player,
     ) {
         val result = dealer vs player
         player.recordResult(result)
         dealer.recordResult(result.inverse)
+        player.addProfit()
+        dealer.addProfit(-player.profit)
```
Wordbe: How about implementing a subtraction operation?

src/main/kotlin/blackjack/model/card/Card.kt
```diff
@@ -1,6 +1,6 @@
 package blackjack.model.card
 
-data class Card(val suit: Suit, val rank: Rank) {
+data class Card(val suit: Suit, val rank: Rank) : Comparable<Card> {
```
Wordbe, line 3: Using Comparable, it's good idea. 👍

src/main/kotlin/blackjack/model/card/Hand.kt
```diff
@@ -1,8 +1,11 @@
 package blackjack.model.card
 
-class Hand(private val dealtCards: MutableList<Card> = mutableListOf()) {
+class Hand(initCards: List<Card> = emptyList()) {
```
Wordbe, line 3: Do we need to keep both initCards and dealtCards? 🙂

src/main/kotlin/blackjack/model/participant/Dealer.kt
```diff
@@ -21,14 +22,14 @@ class Dealer(name: String = "Dealer", internal val deck: Deck = Deck.generateADe
...
-    fun shouldNotStand(): Boolean = _hand.getScore() <= DEALER_STAND
+    fun shouldNotStand(): Boolean = participantHand.getScore() <= DEALER_STAND
 
     fun showAllCards() {
         showAllCards = true
     }
 
-    override fun toString(): String {
-        return name
+    fun addProfit(playerProfit: Chips) {
```
Wordbe: I’d prefer the object name and variable name to be the same — the domain terminology can get a bit confusing otherwise!

src/main/kotlin/blackjack/model/participant/Participant.kt
```diff
@@ -1,29 +1,32 @@
...
+import blackjack.model.Chips
 import blackjack.model.card.Card
 import blackjack.model.card.Hand
 import blackjack.model.result.Result
 import blackjack.model.result.ResultTracker
 
 abstract class Participant(val name: String, protected val resultTracker: ResultTracker) {
     @Suppress("PropertyName")
-    protected val _hand = Hand()
+    protected val participantHand = Hand()
```
Wordbe, line 8: What was the reason for changing hand to participantHand?
Personally, I think the shorter name looks better!

src/main/kotlin/blackjack/model/participant/Player.kt
```diff
@@ -1,11 +1,30 @@
 package blackjack.model.participant
 
+import blackjack.model.Chips
 import blackjack.model.card.Card
 import blackjack.model.result.PlayerResultTracker
+import blackjack.model.result.Result
 
 class Player(name: String) : Participant(name, PlayerResultTracker()) {
+    var bet: Chips = Chips.zero()
```
Wordbe: Is it possible to make bet and profit immutable—ideally without maintaining any internal state?

src/main/kotlin/blackjack/model/result/DealerResultTracker.kt
```diff
@@ -1,6 +1,7 @@
 package blackjack.model.result
 
 class DealerResultTracker : ResultTracker {
+    override var result: Result? = null
```
Wordbe: What was the reason for allowing null?
Could we try refactoring it to avoid allowing null altogether?

src/main/kotlin/blackjack/state/Finished.kt
```diff
@@ -0,0 +1,19 @@
+package blackjack.state
+
+import blackjack.model.card.Card
+
+abstract class Finished : State {
```
Wordbe: Nice 🙂approach.

src/main/kotlin/blackjack/state/Hit.kt
```diff
@@ -0,0 +1,20 @@
+package blackjack.state
+
+import blackjack.model.card.Card
+import blackjack.model.card.Hand
+
+class Hit(override val hand: Hand) : Running() {
+    init {
+        require(hand.size >= 2)
```
Wordbe: Please replace the magic numbers with constants, and update them everywhere else as well.

src/main/kotlin/blackjack/state/Running.kt
```diff
@@ -0,0 +1,7 @@
+package blackjack.state
+
+abstract class Running : State {
+    override fun profit(money: Int): Double {
```
Wordbe: How about creating a Money object to wrap the primitive value? It could improve clarity and type safety. 🙂

src/main/kotlin/blackjack/state/State.kt
```diff
@@ -0,0 +1,14 @@
+package blackjack.state
+
+import blackjack.model.card.Card
+import blackjack.model.card.Hand
+
+interface State {
```
Wordbe: Was there a particular reason you chose an interface over an abstract class? 🙂

Wordbe (changes requested):
Hello Sebas!

Great job on completing the Blackjack mission and submitting your code! 🙌
It was impressive to see how you introduced State to manage the game status of participants.

I've left some comments with points that might be worth considering — please take a look and reflect any changes where possible.

Thank you! 😊

src/main/kotlin/blackjack/model/participant/Participant.kt
SebasNadu: We discussed this with Jason. The reason for renaming _hand to participantHand is to avoid confusion, when you see _hand in a subclass, it immediately looks like a private variable due to the underscore prefix. But it’s actually a protected variable inherited from the parent class. Since it’s not private, using _ can be misleading. Renaming it makes the visibility and origin clearer. We also agreed to reserve the underscore prefix strictly for private variables. Do you agree with this practice, or do you think that both protected and private class could be assigned with _?🤔🤔

src/main/kotlin/blackjack/model/participant/Player.kt
SebasNadu: Yes, I think so, this is a nice idea. I'll check that! Thank you!

src/main/kotlin/blackjack/model/result/DealerResultTracker.kt
SebasNadu: The result property is only relevant for PlayerResultTracker. Initially, I kept it exclusively there. However, this led to a problem: in Participant, which can represent either a player or a dealer, I store a reference to ResultTracker.
When accessing the result in player-specific logic, I had to cast ResultTracker to PlayerResultTracker using as, just to access the result property.
To avoid this unsafe cast, I moved result to the common ResultTracker interface, and for the dealer, it’s simply initialized as null.

An alternative would be to avoid sharing the same ResultTracker type between player and dealer. We could use a PlayerResultTracker in Player and a DealerResultTracker in Dealer, and not reference them via a common Participant field.

I’m open to suggestions. Which do you think is more idiomatic: keeping a shared tracker interface with optional properties, or using distinct tracker fields per subclass?

README.md
SebasNadu: Sorry!🤦

src/main/kotlin/blackjack/model/card/Hand.kt
SebasNadu: I modified the constructor with initCards: List<Card> to improve safety and testability. This allows passing an immutable list when creating a Hand, which is then defensively copied into an internal MutableList (dealtCards).

This way, external code cannot mutate the original input list, and the internal state remains protected. It also makes tests easier to write, since I can initialize a hand with a predefined set of cards without risking accidental mutation of test data.

src/main/kotlin/blackjack/model/participant/Dealer.kt
SebasNadu: Ah, I see! You’re referring to the variable name playerProfit, right? That makes sense, I named it that way because the profit comes from the player. It’s the player’s profit that affects the dealer’s profit. But yes, I get how that can be confusing in the context.

What do you think about renaming it to:
`fun deductFromProfit(chips: Chips)`

I feel this better reflects the actual operation, and avoids any ambiguity about ownership.

SebasNadu:
Hi Wordbe! 👋
Thanks again for your feedback.
In this second iteration, I followed your suggestions: I fixed the issues you pointed out, renamed some methods and variables for clarity, and added the helpful methods you proposed.

Regarding the ResultTracker, I went with the second option to avoid a nullable result. Now, both Player and Dealer contain their own instances of PlayerResultTracker and DealerResultTracker, each with its own record() method. This slightly weakens the benefits of inheritance from Participant, but improves readability and eliminates the nullable concern.

Please ignore the state folder. I experimented with the State pattern but ultimately decided to stick with my initial approach for now. The current implementation doesn’t use any of the classes in that folder. Sorry for the confusion.

Thanks again for your time and insights, you’ve really helped me level up! 🚀🚀

src/main/kotlin/blackjack/model/card/Hand.kt
Wordbe: So you're saying it's used as a `MutableList` only internally - that's a good idea.
Thanks for the explanation! 👍

src/main/kotlin/blackjack/model/participant/Dealer.kt
Wordbe: I'm glad it's clearer now. thank you! 😊

src/main/kotlin/blackjack/model/participant/Participant.kt
Wordbe: Yes, I agree that using `_` only for private properties makes sense. It's a common convention.
But what I was referring to wasn't `_hand`, but `hand` itself!

Ah, I see now. The inherited objects are using `hand`, which makes sense.
Thanks for the clarification! It’s totally fine to keep it as is. Appreciate the explanation!

src/main/kotlin/blackjack/model/result/DealerResultTracker.kt
Wordbe: It seems like having a distinct tracker for each subclass is a good choice.
It improves type safety - no need for casting or null checks
and clearly expresses intent.
Thanks for thinking it through and implementing it! 🙏

src/main/kotlin/blackjack/model/result/PlayerResultTracker.kt
```diff
@@ -1,13 +1,13 @@
 package blackjack.model.result
 
 class PlayerResultTracker : ResultTracker {
-    private var lastResult: Result? = null
+    var result: Result = Result.UNKNOWN
 
     override fun toString(): String {
-        return lastResult?.description ?: "No result recorded."
+        return result?.description ?: "No result recorded."
```
Wordbe: The null operator is no longer needed.
Also, if `toString()` is being used for the View, consider moving that responsibility to the View layer instead. (Same feedback as before) 🙂

src/main/kotlin/blackjack/model/participant/Participants.kt
```diff
@@ -1,15 +1,23 @@
...
 
+import blackjack.model.Chips
+
 class Participants(val dealer: Dealer, val players: List<Player>) {
     val dealerScore: Int
         get() = dealer.getScore()
     val dealerName: String
         get() = dealer.name
     val dealerResult: String
         get() = dealer.result
```
Wordbe: Since it's no longer used, please remove it!

Wordbe (changes requested):
Hello Sebas!

Thank you for your great work on Blackjack mission Step 2.
Are you planning to merge the part you created with `State` into the main code? Please check on this.

Also, please review the comments I left and apply any changes you can.
Thanks! 😊

SebasNadu:
Yes, you're right. I'll create a new branch for the upcoming state feature and keep the step2 branch clean and ready for merging. Thanks a lot for your insights and the patience.

SebasNadu:
Hi Wordbe! 👋

I’ve removed the unused state folder from this branch, cleaned up the unused result property in Participants, and eliminated the unnecessary null check in PlayerResultTracker.

Thanks again for your thoughtful review, and for your patience with the small oversights. I’ll make sure to pay closer attention to those details going forward 🙏🙂

I hope the code is now ready to be merged! 🚀

Wordbe (approved):
Hi Sebas!
Thanks a lot for the cleanups. I really appreciate the attention to detail.
The code looks great now, and is in good shape for merging.

As you've already started doing, I really hope you take the time to apply the State pattern on your own. It's a great way to truly experience object-oriented design and the power of polymorphism.

Well done so far, and best of luck with your next mission! You've got this! 💪🚀

PR #20, opened 2025-07-14
[Blackjack - Step2] Submit Deniz' mission

deniz-oezdemir:
Hi Eugine,

Thanks for reviewing step1. I’ve made the changes you requested in step2. This PR is ready for review.

I have two questions regarding the same topic:

- Specifically, should I removed the result calculation and display logic from step1 completely since it is no longer used in step2?
For now, I have only commented out the method calls in `Controller` but left the `displayResults` method in `OutputView` although it is not used anymore.

- Generally, is it best practice to delete all unused code?

Also, I am always looking to improve my code to be more Kotlin-idiomatic. I would appreciate any feedback on that.

Thanks in advance for your time. 🙏 

Best,
Deniz


deniz-oezdemir:
Closed due to merge conflicts, will open a new PR.

PR #21, opened 2025-07-14
[Blackjack - Step2] Submit Deniz' mission

deniz-oezdemir:
Hi Eugine,

Thanks for reviewing step1. I have made the changes you recommended in step2. This PR is ready for review.

I have two questions regarding the same topic:
- Specifically, should I removed the result calculation and display logic from step1 completely since it is no longer used in step2? For now, I have only commented out the method calls in `Controller` but left the `displayResults` method in `OutputView` although it is not used anymore.
- Generally, is it best practice to delete all unused code?

Also, I am always looking to improve my code to be more Kotlin-idiomatic. I would appreciate any feedback on that.

Thanks in advance for your time. 🙏

Best, Deniz


src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -30,9 +29,13 @@ class Controller {
...
-        val evaluator = ResultEvaluator(players, dealer)
-        val results = evaluator.calculateResults()
-        OutputView.displayResults(results)
+        /* step1 calculations
+         * val evaluator = ResultEvaluator(players, dealer)
+         * val results = evaluator.calculateResults()
+         * OutputView.displayResults(results)
+         */
+        dealer.settleRound()
+        OutputView.displayEarnings(dealer, players)
```
pkeugine: Thanks for showing me the difference!
The thing is, I can see the difference when I check the 'changed' code in GitHub:

Screenshot 2025-07-15 at 9 14 58 PM

So I think we can safely remove this comment before this PR gets merged 😃

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -42,7 +45,11 @@ class Controller {
 
     private fun initializePlayers(): Players {
         val playerNames = InputView.readNames()
-        return Players(playerNames.map { Player(it) })
+        val playerBets = InputView.readBets(playerNames)
+
+        val players = playerNames.zip(playerBets) { name, bet -> Player(name = name, bet = bet) }
```
pkeugine, line 45: Nice use of `zip` 👍

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -13,11 +15,18 @@ abstract class Participant {
...
+        earning = when {
+            isBusted()                                                  -> -bet
+            dealer.isBusted()                                           -> +bet
+            handCards.isBlackjack() && dealer.handCards.isBlackjack()   -> 0
+            handCards.isBlackjack()                                     -> (bet * 1.5).toInt()
+            dealer.handCards.isBlackjack()                              -> -bet
+            handCards.total > dealer.handCards.total                    -> +bet
+            handCards.total < dealer.handCards.total                    -> -bet
+            else                                                        -> 0
+        }
```
pkeugine: Wow, I've never thought of using when like this; it looks pretty good.
A good way to make it easier to understand quickly might be adding some empty lines to group similar ones :  
```suggestion
    fun settlePlayerAndDealer(dealer: Dealer) {
        earning = when {
            isBusted()                                                  -> -bet
            dealer.isBusted()                                           -> +bet
            
            handCards.isBlackjack() && dealer.handCards.isBlackjack()   -> 0
            handCards.isBlackjack()                                     -> (bet * 1.5).toInt()
            dealer.handCards.isBlackjack()                              -> -bet
            
            handCards.total > dealer.handCards.total                    -> +bet
            handCards.total < dealer.handCards.total                    -> -bet
            
            else                                                        -> 0
        }
```

But this is just a small suggestion, so please do as you please 😉

src/main/kotlin/study/DSL.kt
```diff
@@ -11,16 +11,17 @@ data class Person(
     val name: String,
     val company: String?,
     val skills: Skills,
-    val languages: List<Language>
+    val languages: List<Language>,
```
pkeugine, line 14: Nice addition of trailing comma 👍

src/main/kotlin/study/DSL.kt
```diff
@@ -11,16 +11,17 @@ data class Person(
...
 
 data class Skills(
     val soft: List<String>,
-    val hard: List<String>
+    val hard: List<String>,
 )
 
 data class Language(val name: String, val level: Int)
 
+@PersonDSL
```
pkeugine, line 24: Nice work of applying previous comment about `@DslMarker`! 💯

src/test/kotlin/blackjack/DeckTest.kt
```diff
@@ -29,12 +26,11 @@ class DeckTest {
     }
 
     @Test
-    fun `test draw more than deck`() {
+    fun `throws when drawing more cards than deck contains`() {
```
pkeugine, line 29: This test name definitely describes the test much better than the previous one 👍

src/test/kotlin/blackjack/ParticipantTest.kt
```diff
@@ -0,0 +1,143 @@
...
+import blackjack.model.HandCards
+import blackjack.model.Player
+import blackjack.model.Players
+import blackjack.model.Rank
+import blackjack.model.Suit
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class ParticipantTest {
+    private val deck = CardDeck()
```
pkeugine: Since this `deck` is only used in one test, why don't we keep it in the test scope?

Also, I think this would be a great time to think about tests, shared fixture, and test isolation.
If we were to use this `deck` in any other tests, though it might look okay for now, there will me moments where a test modifies this `deck` object and affect other tests. This will lead to unexpected test results, such as tests failing when they shouldn't or even worse, tests succeeding when they should fail.

Having tests create fixtures by themselves, or re-initiating shared fixtures are very important tasks when it comes to maintaining tests. It'll be great to think about this point in later missions :)

pkeugine (changes requested):
Hello Deniz 👍 
The PR looks very solid, and I think we can wrap up this mission right away.
Most of the comments are just me verifying past comments and saying 'nice job'!
However, I thought you might want to edit some small parts,
so I'll keep it open, and merge it when you request a review once again.

If you have any questions or thoughts, feel free to DM me or leave comments here.
And as always,
Happy Hacking! 👏 👏 👏

src/main/kotlin/blackjack/controller/Controller.kt
deniz-oezdemir: Thanks for the hint, I will remove it. :-)

src/main/kotlin/blackjack/model/Participant.kt
deniz-oezdemir: Yes, I agree. Grouping the cases makes it more readable. 😄

src/main/kotlin/study/DSL.kt
deniz-oezdemir: Thanks, but the credit goes to the auto-formatter here. 😉

src/test/kotlin/blackjack/ParticipantTest.kt
deniz-oezdemir: Yes, the deck can definitely be instantiated in the respective test for this case. I will refactor it accordingly.

That's true, it will definitely not be good if the tests interact with each other because they use the same underlying data. For the future I will keep possible test interference in mind. Thanks for the hint! 🙏

deniz-oezdemir:
Hi @pkeugine ! 👋

Thanks so much for the kind words and thoughtful review! 😊

I went ahead and implemented the three changes you suggested, and everything should be good to go now.

Feel free to give it another look whenever you have a moment. It's ready for review.

Thanks again, and happy hacking to you too! 🙌

Best,
Deniz

README.md
```diff
@@ -50,14 +50,17 @@ jason: Lose
   - [x] comma separated
```
pkeugine, line 50: **Q. Generally, is it best practice to delete all unused code?**

Personally, yes, I do think it is a good practice to remove unused code.
Even if we remove it, it'll still be in our git log, so it's not 100% gone from our work so we don't have to worry about it vanishing completely. Also, keeping unused code in the project tends to distract teams from focusing on what's actually important, and might confuse us by making us think 'this might be part of the logic we don't know', 'maybe it's important but we can't keep track of it'. When we need it again, it's not that hard to put it in.

pkeugine (approved):
Hello Deniz :)
Nice job on step 2, and I've answered the question you asked in the PR message.
It's my personal opinion based off of experience, so the answer might vary from developer to developer.

Anyways, the project looks like it's ready to get merged!
I hope you've learned a lot during this mission, and do well on your next mission as well.
Wish you all the luck, and as always,
Happy Hacking! 😄 😄 😄

PR #22, opened 2025-07-14
[Blackjack - Step2] Submit Arekiu's mission

arekiu:
## Add new features to Blackjack project

- Create functionality to integrate players bet and calculate earnings after comparing final results.

### Notes about implementation

- In the subject, the only mentioned requirements about bet are:
    - If a player draws a card and the total exceeds 21, they lose their entire bet.
    - If a player hits 21 with the initial two cards (Blackjack), they receive 1.5x their bet.
    - If both the player and dealer have Blackjack, the player’s bet is returned.

- There is no mention about a tie (when is not Blackjack), so I decided that a tie return 0 earnings.
- Dealer win or lose is calculated depending on players win or lose.

Thank you so much for your time to review my code. Have a nice day!

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ The goal is to get as close to 21 as possible without going over.
 ## Step 2 new features
 
 - Player bet
-- [ ] Get bet from every player
+- [x] Get bet from every player
 - [ ] If initial hand is equal 21, player wins 1.5 bet
 - [ ] If initial hand is equal 21, but dealer has also 21. Bet is returned
 - [ ] If player is busted, loses their bet.
--- a/README.md
+++ b/README.md
@@ -56,8 +56,8 @@ The goal is to get as close to 21 as possible without going over.
 - [x] If player is busted, loses their bet.
 
 ## Example output
-
-`Enter the names of the players (comma-separated):
+```
+Enter the names of the players (comma-separated):
 pobi,jason
 
 Enter pobi’s betting amount:
@@ -89,4 +89,5 @@ jason's cards: 7♣, K♠ – Total: 17
 \## Final Earnings
 Dealer: 10000  
 pobi: 10000  
-jason: -20000`
+jason: -20000
+```
--- a/README.md
+++ b/README.md
@@ -51,9 +51,9 @@ The goal is to get as close to 21 as possible without going over.
 
 - Player bet
 - [x] Get bet from every player
-- [ ] If initial hand is equal 21, player wins 1.5 bet
-- [ ] If initial hand is equal 21, but dealer has also 21. Bet is returned
-- [ ] If player is busted, loses their bet.
+- [x] If initial hand is equal 21, player wins 1.5 bet
+- [x] If initial hand is equal 21, but dealer has also 21. Bet is returned
+- [x] If player is busted, loses their bet.
 
 ## Example output
 
--- a/README.md
+++ b/README.md
@@ -46,3 +46,47 @@ The goal is to get as close to 21 as possible without going over.
 - [x] Display dealer message (if 16 or less)
 - [x] Display all the cards with total value
 - [x] Display Winning Result
+
+## Step 2 new features
+
+- Player bet
+- [ ] Get bet from every player
+- [ ] If initial hand is equal 21, player wins 1.5 bet
+- [ ] If initial hand is equal 21, but dealer has also 21. Bet is returned
+- [ ] If player is busted, loses their bet.
+
+## Example output
+
+`Enter the names of the players (comma-separated):
+pobi,jason
+
+Enter pobi’s betting amount:
+10000
+
+Enter jason’s betting amount:
+20000
+
+Dealing two cards to dealer, pobi, jason.
+Dealer: 3♦  
+pobi's cards: 2♥, 8♠  
+jason's cards: 7♣, K♠
+
+Would pobi like to draw another card? (y for yes, n for no)  
+y  
+pobi's cards: 2♥, 8♠, A♣  
+Would pobi like to draw another card? (y for yes, n for no)  
+n  
+Would jason like to draw another card? (y for yes, n for no)  
+n  
+jason's cards: 7♣, K♠
+
+Dealer draws one more card due to having 16 or less.
+
+Dealer’s cards: 3♦, 9♣, 8♦ – Total: 20  
+pobi's cards: 2♥, 8♠, A♣ – Total: 21  
+jason's cards: 7♣, K♠ – Total: 17
+
+\## Final Earnings
+Dealer: 10000  
+pobi: 10000  
+jason: -20000`
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -46,7 +46,7 @@ class Game(
                 val answer = InputView.askToHit(it.name)
                 if (answer) {
                     dealer.dealCardToPlayer(it)
-                    it.updatePlayingStatus(it.cardsInHand.checkIfIsBustHand())
+                    it.updatePlayingStatus(it.cardsInHand.isBustHand())
                     println(OutputView.showHandCards(it, false))
                 }
             } while (answer && it.isPlaying)
@@ -60,7 +60,7 @@ class Game(
             dealer.selfDrawCard()
             mustDraw = dealer.mustDraw(dealer.cardsInHand.calculateTotalValueOfCards())
         }
-        dealer.updatePlayingStatus(dealer.cardsInHand.checkIfIsBustHand())
+        dealer.updatePlayingStatus(dealer.cardsInHand.isBustHand())
     }
 
     private fun compareFinalCards(players: List<Player>) {
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -12,11 +12,19 @@ class Game(
         val playersName = InputView.askPlayerNames()
         val players = playersName.map { Player(it) }
 
+        collectBets(players)
+
         assignInitialCards(players)
 
         hitOrStay(players)
 
         compareFinalCards(players)
+    }
+
+    private fun collectBets(players: List<Player>) {
+        players.forEach { player ->
+            player.setBet(InputView.askPlayerBet(player.name))
+        }
     }
 
     private fun assignInitialCards(players: List<Player>) {
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -25,7 +25,7 @@ class Game(
 
     private fun collectBets(players: Players) {
         players.values.forEach { player ->
-            player.setBet(InputView.askPlayerBet(player.name))
+            player.placeBet(InputView.askPlayerBet(player.name))
         }
     }
 
@@ -70,7 +70,7 @@ class Game(
 
     private fun calculateEarnings(players: Players) {
         players.values
-            .forEach { player -> player.updateEarnings(ResultCalculation.calculatePlayerResult(player, dealer)) }
-        dealer.updateEarnings(players.calculateTotalPlayersEarning())
+            .forEach { player -> player.receiveEarningsOrLoses(ResultCalculation.calculatePlayerResult(player, dealer)) }
+        dealer.calculateBalance(players.calculateTotalPlayersEarning())
     }
 }
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -48,7 +48,7 @@ class Game(
                 if (answer) {
                     dealer.dealCardToPlayer(it)
                     it.updatePlayingStatus(it.cardsInHand.isBustHand())
-                    println(OutputView.showHandCards(it, false))
+                    OutputView.displayCardsAfterHit(it)
                 }
             } while (answer && it.isPlaying)
         }
--- a/src/main/kotlin/blackjack/controller/Game.kt
+++ b/src/main/kotlin/blackjack/controller/Game.kt
@@ -24,7 +24,7 @@ class Game(
     }
 
     private fun collectBets(players: Players) {
-        players.getPlayers().forEach { player ->
+        players.members.forEach { player ->
             player.setBet(InputView.askPlayerBet(player.name))
         }
     }
@@ -39,11 +39,11 @@ class Game(
         askPlayersToHit(players)
         dealerDraws()
         OutputView.displayDealersCardsWithTotalValue(dealer)
-        players.getPlayers().forEach { player -> OutputView.displayCardsWithTotalValue(player) }
+        players.members.forEach { player -> OutputView.displayCardsWithTotalValue(player) }
     }
 
     private fun askPlayersToHit(players: Players) {
-        players.getPlayers().forEach {
+        players.members.forEach {
             do {
                 val answer = InputView.askToHit(it.name)
                 if (answer) {
@@ -68,7 +68,7 @@ class Game(
     private fun compareFinalCards(players: Players) {
         if (dealer.isPlaying) {
             val dealerPoints = dealer.cardsInHand.calculateTotalValueOfCards()
-            players.getPlayers().forEach {
+            players.members.forEach {
                 it.updatePlayingStatus(it.hasLessPointsThanDealer(dealerPoints))
             }
         }
@@ -77,7 +77,7 @@ class Game(
     }
 
     private fun calculateEarnings(players: Players) {
-        players.getPlayers()
+        players.members
             .forEach { player -> player.updateEarnings(ResultCalculation.calculatePlayerEarnings(player, dealer)) }
         dealer.updateEarnings(ResultCalculation.calculateDealerEarnings(dealer, players))
     }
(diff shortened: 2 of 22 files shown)


README.md
```diff
@@ -46,3 +46,48 @@ The goal is to get as close to 21 as possible without going over.
...
 - [x] Display all the cards with total value
 - [x] Display Winning Result
+
+## Step 2 new features
+
+- Player bet
+- [x] Get bet from every player
+- [x] If initial hand is equal 21, player wins 1.5 bet
+- [x] If initial hand is equal 21, but dealer has also 21. Bet is returned
+- [x] If player is busted, loses their bet.
```
Hyeon9mak, line 56: 👍

src/main/kotlin/blackjack/model/Cards.kt
```diff
@@ -17,4 +17,8 @@ class Cards(val cards: MutableList<Card>) {
         }
         return totalValueOfCards
     }
+
+    fun hasBlackJack() = cards.size == 2 && calculateTotalValueOfCards() == BlackJackValues.BLACK_JACK
+
```
Hyeon9mak, line 21: yeah! this is the REAL Blackjack 😄 👍

src/main/kotlin/blackjack/model/Cards.kt
```diff
@@ -17,4 +17,8 @@ class Cards(val cards: MutableList<Card>) {
...
+    fun isBustHand() = calculateTotalValueOfCards() > BlackJackValues.HAND_VALUE_LIMIT
```
Hyeon9mak, line 22: very clear 👍 💯

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -1,23 +1,24 @@
...
 
     fun drawCard(card: Card) {
         cardsInHand.addCard(card)
     }
 
-    fun updateBustedStatus() {
-            isBusted = true
+    fun updatePlayingStatus(condition: Boolean) {
+        if (condition)
+            isPlaying = false
```
Hyeon9mak: It seems like the `isPlaying` flag is currently used mainly for display purposes, rather than holding actual logic.
How about asking the Hand in real time whether it's busted?
That way, you might be able to remove the `isPlaying` field 😄

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -1,6 +1,6 @@
 package blackjack.model
 
-class Dealer(name: String = "Dealer") : Participant(name) {
+class Dealer() : Participant() {
```
Hyeon9mak:
```suggestion
class Dealer : Participant() {
```

😉

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -13,68 +13,81 @@ object OutputView {
         return "${face}${symbol}"
     }
 
-    fun showHandCards(participant: Participant, isFirstRound: Boolean): String {
-        return when {
-            participant is Dealer && isFirstRound -> "${participant.name}: ${showCardFaceAndSymbol(participant.cardsInHand.cards[0])}"
+    fun showDealerFirstHandCards(dealer: Dealer): String {
+        return ("Dealer: ${showCardFaceAndSymbol(dealer.cardsInHand.cards[0])}")
+    }
```
Hyeon9mak, line 17: Finally! Now we can focus on the game without even knowing the dealer’s name!
And if we ever want to change how the dealer appears, we only need to look at the view 👍

src/main/kotlin/blackjack/model/Players.kt
```diff
@@ -0,0 +1,12 @@
+package blackjack.model
+
+class Players(private val players: List<Player>) {
+    init {
+        require(players.isNotEmpty()) { "Player list must not be empty" }
+        require(players.size <= 7) { "Maximum number of players must be 7" }
+    }
+
+    fun getPlayers() = players
+
```
Hyeon9mak:
```suggestion
class Players(
    val players: List<Player>,
) {
    init {
        require(players.isNotEmpty()) { "Player list must not be empty" }
        require(players.size <= 7) { "Maximum number of players must be 7" }
    }
```
same code 😉

src/main/kotlin/blackjack/model/ResultCalculation.kt
```diff
@@ -0,0 +1,34 @@
...
+    private fun hasPlayerBJ(player: Player) = player.cardsInHand.hasBlackJack()
+
+    private fun hasAnyPlayerBJ(players: Players) = players.getPlayers().any { player -> player.cardsInHand.hasBlackJack() }
+
+    fun calculatePlayerEarnings(player: Player, dealer: Dealer): Int {
+        return when {
+            hasDealerBJ(dealer) && hasPlayerBJ(player) -> 0
+            hasPlayerBJ(player) -> (player.bet * BlackJackValues.BET_MULTIPLICATOR).toInt()
+            player.isPlaying && !dealer.isPlaying -> player.bet
+            !player.isPlaying -> -(player.bet)
```
Hyeon9mak: u can append new property for Participants!
it's more readable, isn't it? 😄 

```kotlin
var isPlaying: Boolean = true
    private set

val isBusted: Boolean = !isPlaying
```

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -16,6 +16,19 @@ object InputView {
...
 
+    fun askPlayerBet(name: String): Int {
+        println("Enter $name’s betting amount: ")
+        try {
+            val bet =
+                readln()
+                    .toInt()
+            require(bet > 0) { "Bet must be a positive number." }
+            return bet
+        } catch (e: IllegalArgumentException) {
```
Hyeon9mak: Making sure the bet amount is always positive could be seen as a responsibility of the Model.
But on the other hand, it’s also such a common-sense rule that handling it in the View could be reasonable too. 👍 

For cases like this where the boundary is a bit fuzzy,
it’s best to discuss with your teammates and align on the approach! 😄

src/test/kotlin/blackjack/model/CardsTest.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun `should return false when cards in hand dont fulfil condition for BlackJack`() {
+        val cardsInHand = Cards(
+            mutableListOf(
+                Card(Rank.ACE, Suit.HEARTS),
+                Card(Rank.NINE, Suit.DIAMONDS),
+                Card(Rank.ACE, Suit.DIAMONDS)
+            )
+        )
+        assertThat(cardsInHand.hasBlackJack()).isFalse()
+    }
```
Hyeon9mak: nice test! 👍 

How about explaining it in a way that more directly reflects the real-world rule?

- 'If it's 21 points with exactly 2 cards, it's Blackjack.'
- 'Even if it's 21 points, it's not Blackjack unless there are exactly 2 cards.'

Hyeon9mak (changes requested):
Great work in step2 Alex! 👍 👍 

I think the responsibilities between the `Model` and the `View` are mostly well separated.
Game does look a bit complex, but that seems understandable given that Blackjack involves continuous user input. 😄 

I've left a few minor comments — feel free to ask if you have any questions!

src/main/kotlin/blackjack/model/Dealer.kt
arekiu: clear!

src/main/kotlin/blackjack/view/OutputView.kt
arekiu: Finally I got your point! 😄

src/main/kotlin/blackjack/model/Players.kt
arekiu: Yes, that getter was extra. I guess that I did it because it was looking weird using players.players and I don´t have a better idea to name the Class or the property 😢 

Currently looks like players.members

src/test/kotlin/blackjack/model/CardsTest.kt
arekiu: It has much sense! Change applied

src/main/kotlin/blackjack/model/Participant.kt
arekiu: Right! I removed it and I added a method isAbleToHit() in Cards class. So together with isBustHand() , I can check in real time if the player can draw or not, or if it´s out of the game.

Note: I know that isAbleToHit is basically !isBustHand, but I wrote the new method for better readability.

src/main/kotlin/blackjack/model/ResultCalculation.kt
arekiu: I decided to delete isPlaying and now there´s a new logic to check this (more details in the changes you asked me in Participant.kt)

src/main/kotlin/blackjack/view/InputView.kt
arekiu: Right! I decided to move this check to setBet() in Player class. I think now is more accurate

arekiu:
Good morning/afternoon/evening! 

I applied all the requested changes and I see already how the code looks much better!
Thank you so much for your time to review my project 😄

src/main/kotlin/blackjack/controller/Game.kt
```diff
@@ -19,49 +23,54 @@ class Game(
...
             do {
                 val answer = InputView.askToHit(it.name)
                 if (answer) {
                     dealer.dealCardToPlayer(it)
-                    it.checkCardsValueLimit()
-                    println(OutputView.showHandCards(it, false))
+                    OutputView.displayCardsAfterHit(it)
                 }
-            } while (answer && !it.isBusted)
+            } while (answer && it.cardsInHand.isAbleToHit())
```
Hyeon9mak: For call the `isAbleToHit()`, player need call getter method.

> Don't use getter. Send message to Object.

How about this?

```kotlin
abstract class Participant(
    val cardsInHand: Cards = Cards(mutableListOf())
) {
    fun isAbleToHit(): Boolean {
        return cardsInHand.isAbleToHit()
    }
```

Hyeon9mak: > Note: I know that isAbleToHit is basically !isBustHand, but I wrote the new method for better readability.

💯 💯 💯 
I always left comment like this
"How about remake negative method? `!` operator make human error easier."

src/main/kotlin/blackjack/model/Players.kt
```diff
@@ -0,0 +1,12 @@
+package blackjack.model
+
+class Players(
+    val members: List<Player>
+) {
+    init {
```
Hyeon9mak: > I don´t have a better idea to name the Class or the property 😢

Naming variables is always one of the hardest parts. I totally get that 😄 
It gets even trickier with value objects like `Players`, where the class name itself already carries meaning.

That’s why I often prefer using names like `value` or `values`.
e.g. `players.values`, `money.value`

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -1,12 +1,13 @@
...
+    var bet: Int = 0
+        private set
 
-    fun comparePointsAgainstDealer(dealerPoints: Int) {
-        if (cardsInHand.calculateTotalValueOfCards() <= dealerPoints) {
-            updateBustedStatus()
-        }
+    fun setBet(bet: Int) {
+        require(bet > 0) { "Bet must be a positive number." }
+        this.bet = bet
```
Hyeon9mak: 👍 
or, u can create new value class 😉 
and u can generate new test code!

Hyeon9mak (changes requested):
Did a great job applying the feedback quickly Alex!
All the requirements are met, so there’s not much that needs to be changed.

I heard Jason recently introduced the State pattern —
I figured you might have some questions about it or be interested in trying it out,
so I left a few very simple comments along with a final 'Request Changes'.

If you don’t have any major questions, feel free to request a review again right away!

src/main/kotlin/blackjack/controller/Game.kt
arekiu: Clear! I just named the method in Cards as "isLowHand" (lower than 21) and that will determine the result of isAbleToHit in Participant class

src/main/kotlin/blackjack/model/Players.kt
arekiu: Clear! I´m using values

src/main/kotlin/blackjack/model/Player.kt
arekiu: Okk I created playerWallet class to store information about the bet and balance

arekiu:
Thank you so much for your review!
I´m still trying to understand the use of State Pattern, so I tried a minimalist application (PlayerState) to track if the player will be able to draw a next card. After trying this, I see all the potential benefit of it. For example it could make the results calculation much easier, with tracking when the player ended their participation in an specific stage of the game. I can keep working on more features with state, in the meanwhile, let me know if my implementation is correct.

Also, I add to the project a little task to practice DSL that we saw in class. I would like to know, mainly, if the creation of skills and languages inside PersonBuilder are right.

Thank you so much for your time! Have a nice day!

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -1,12 +1,28 @@
 package blackjack.model
 
 class Player(
-    name: String,
-) : Participant(name) {
+    val name: String,
+) : Participant() {
+    val wallet = PlayerWallet()
+    var state: PlayerState = Playing
```
Hyeon9mak, line 7: niiice 💯

src/test/kotlin/blackjack/model/PlayerWalletTest.kt
```diff
@@ -0,0 +1,27 @@
...
+        val wallet = PlayerWallet()
+
+        assertThatThrownBy { wallet.placeBet(0) }
+            .isInstanceOf(IllegalArgumentException::class.java)
+            .hasMessageContaining("Bet must be a positive number.")
+
+        assertThatThrownBy { wallet.placeBet(-50) }
+            .isInstanceOf(IllegalArgumentException::class.java)
+            .hasMessageContaining("Bet must be a positive number.")
+    }
```
Hyeon9mak, line 26: 😉 👍 💯

src/main/kotlin/blackjack/model/PlayerState.kt
```diff
@@ -0,0 +1,17 @@
...
+    override fun canHit() = true
+}
+
+object Busted : PlayerState {
+    override fun canHit() = false
+}
+
+object Blackjack : PlayerState {
+    override fun canHit() = false
+}
```
Hyeon9mak, line 17: If you take it one step further, instead of asking the state "Can you transition?", you can command it directly — like "Proceed to the next state."

For example, when you tell Playing to "draw a card", if the score is fine, it stays in Playing.

But if the draw goes badly, it transitions into Busted.
And if you tell Busted to "draw a card", it could throw an IllegalStateException.

All the conditional logic is pushed inside each State object.

The State pattern reflects the desire to isolate the rules of each state, so we can just open one class file when making changes. 😄

src/test/kotlin/DSLTest.kt
```diff
@@ -0,0 +1,126 @@
...
+
+    fun skills(block: SkillBuilder.() -> Unit) {
+        skills = SkillBuilder().apply(block).build()
+    }
+
+    fun languages(block: LanguageBuilder.() -> Unit) {
+        languages = LanguageBuilder().apply(block).build()
+    }
+
+    fun build(): Person = Person(name, company, skills, languages)
```
Hyeon9mak, line 70: Normally, if a property isn’t nullable and we forget to initialize something like name when creating a Person via a builder, it would throw an error.

But since you've made the property nullable, it works even if not all fields are filled — well done!

src/test/kotlin/DSLTest.kt
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    fun build(): Skill = Skill(soft, hard)
+}
+
+class PersonBuilder {
+    private var name: String? = null
+    private var company: String? = null
+    private var skills: Skill? = null
+    private var languages: List<Language>? = emptyList()
```
Hyeon9mak, line 52: I think `languages` no need `?` operator 😄

src/test/kotlin/DSLTest.kt
```diff
@@ -0,0 +1,126 @@
...
+        hard.add(value)
+    }
+
+    fun build(): Skill = Skill(soft, hard)
+}
+
+class PersonBuilder {
+    private var name: String? = null
+    private var company: String? = null
+    private var skills: Skill? = null
```
Hyeon9mak, line 51: the best thing is, give default value for `Skill` enum class.
Because an enum class is meant to represent all possible constants or states by itself, making it nullable is essentially like adding an extra state — null.

Hyeon9mak (approved):
Hi Alex!
You reflected the feedback quickly and nicely.

It was also great to see you experimenting with Kotlin DSL and the State Pattern — even if lightly. 
There are still some areas that could be improved, but I believe you'll have plenty of opportunities to revisit them after completing all the missions, so I won’t hold things up here. 😉 

You've met all the requirements, so I'm approving this PR! 👍 
Really great work. Hope you enjoy the next missions too!

PR #27, opened 2025-07-15
[Blackjack - Step2] Submit Vito's mission

VittorioDeMarzi:
Hi   🙂
First of all, thank you so much for taking the time to review my code, I really appreciate your feedback and support 🙏

After pretty much wiping out everything from Step 1, I decided to rebuild the game from scratch using a State Pattern approach.

This PR introduces the core Blackjack game logic and structure. The goal was to model the game as a finite state machine where each Participant (Player or Dealer) transitions through specific states based on standard Blackjack rules.

## State Pattern 
I defined a State interface with the main transitions:
- draw(),
- stay(),
- profit() (for payout calculation).

The states are divided as follows:
- Running
   - FirstTurn: handles the first 2 cards
   - Hit: for drawing additional cards
- Finished
   - Blackjack: 2 cards that total 21
   - Stay: player chooses to stop
   - Bust: total over 21

Each state encapsulates its own transition logic and constraints. 
For example, you can't stay() in FirstTurn unless 2 cards are present.

## Card Refactoring: Object Pool
I also refactored the Card class to use an object pool, to ensure memory efficiency:
 - Only 52 unique Card instances are created.
 - Cards are reused across all decks.

## Participant Abstraction
I introduced a Participant abstraction to share logic between Player and Dealer.
It exposes the current State and delegates transitions to the State object.

## Betting Logic
Each player places a bet at the beginning of the game.

## Questions for Review
- BlackjackGame currently initializes the hands directly in its init block. Would it be cleaner to move that logic into a separate method?
- I'm passing askForCard() as a function to handle user interaction. Do you think this approach is clean and flexible enough? Or would you recommend a different abstraction?

Thanks again for your feedback! I'm open to any suggestions you might have to improve the structure or clarity of the implementation.


src/main/kotlin/blackjack/controller/BlackJackController.kt
```diff
@@ -1,27 +1,40 @@
...
-import blackjack.model.Statistics
+import blackjack.model.BlackjackGame
+import blackjack.model.GameLogic
+import blackjack.model.ParticipantsFactory
 import blackjack.view.InputView
 import blackjack.view.OutputView
 
-object BlackJackController {
-    fun play() {
+class BlackJackController() {
```
Wordbe: You don't need to define a default constructor. It's automatically generated!

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -1,21 +1,20 @@
...
+    fun stay() {
+        state = state.stay()
+    }
 
-    fun numberInHand() = hand.numberOfCards()
-
-    fun calculatePoints() = hand.calculatePoints()
+    fun setBettingAmount(bettingAmount: Int) {
+        this.bettingAmount = bettingAmount
+    }
```
Wordbe: Instead of manually writing setters and getters, consider removing `private` so they can be generated automatically — it’ll be simpler and cleaner! 🙂

src/main/kotlin/blackjack/model/BlackjackGame.kt
```diff
@@ -0,0 +1,36 @@
...
+    val cardDeck = CardDeck()
+
+    init {
+        setUpTable()
+    }
+
+    fun setUpTable() {
+        players.forEach { player ->
+            repeat(2) {
+                player.state = player.state.draw(cardDeck.drawCard())
```
Wordbe: Try handling the state changes inside the `Player` object! That way, the logic stays encapsulated and cohesive.

src/main/kotlin/blackjack/model/BlackjackGame.kt
```diff
@@ -0,0 +1,36 @@
...
+            }
+        }
+        repeat(2) { dealer.state = dealer.state.draw(cardDeck.drawCard()) }
```
Wordbe: Here as well. please apply the same change!

src/main/kotlin/blackjack/model/BlackjackGame.kt
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    fun play() {
+        dealer.playTurn(cardDeck)
+        players.forEach {
+            while (it.state is Running && askForCard(it)) {
```
Wordbe: Instead of using `is`, how about leveraging polymorphism? It can lead to cleaner and more maintainable code. 🙂

src/main/kotlin/blackjack/model/Card.kt
```diff
@@ -1,3 +1,27 @@
 package blackjack.model
 
-data class Card(val suit: Suit, val rank: Rank)
+import blackjack.view.Errors
+
+data class Card(val suit: Suit, val rank: Rank) {
```
Wordbe, line 5: Since all instances are created in advance, it seems like identity can be guaranteed when `(suit, rank)` are the same.

In that case, what was your reason for using a `data class`? What purpose did it serve here? 🙂

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
+package blackjack.model
+
+import blackjack.states.FirstTurn
+import blackjack.states.Running
+
+class Dealer() : Participant("Dealer", FirstTurn(Hand(emptyList()))) {
+    fun playTurn(deck: CardDeck) {
+        while (state is Running && state.hand.calculatePoints() < 17) {
```
Wordbe: Please replace the magic numbers with constants. 🙂

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,15 @@
...
+            state = state.draw(deck.drawCard())
+        }
+        if (state is Running) {
```
Wordbe: Try removing the `is` check once and see how it goes!

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -1,19 +1,14 @@
...
-        get() = hold.cards
-
-    fun numberOfCards() = hold.cards.size
-
-    fun addCard(card: Card) = this.hold.add(card)
+    operator fun plus(card: Card): Hand = Hand(this.cards + card)
 
     fun calculatePoints(): Int {
-        val countOfAce = hold.cards.filter { it.rank.value == 1 }.size
+        val countOfAce = cards.filter { it.rank.value == 1 }.size
```
Wordbe: Instead of comparing values, it might be better to compare the enum itself directly.

src/main/kotlin/blackjack/states/FirstTurn.kt
```diff
@@ -0,0 +1,26 @@
+package blackjack.states
+import blackjack.model.Card
+import blackjack.model.Hand
+import blackjack.view.Errors
+
+class FirstTurn(override val hand: Hand = Hand(emptyList())) : Running {
+    override fun draw(card: Card): State {
+        val hand = this.hand + card
+        if (hand.size == 2) {
+            if (hand.calculatePoints() == 21) return Blackjack(hand)
```
Wordbe: Please replace the magic numbers with constants. 🙂

src/main/kotlin/blackjack/states/FirstTurn.kt
```diff
@@ -0,0 +1,26 @@
...
+            return Hit(hand)
+        }
+        return FirstTurn(hand)
+    }
+
+    override fun stay(): State {
+        throw IllegalStateException(Errors.INVALID_STAY.message)
+    }
+
+    override fun profit(
```
Wordbe: The `profit` method looks like it could be handled commonly within the `Running` state.
That might simplify your logic!

src/main/kotlin/blackjack/states/State.kt
```diff
@@ -0,0 +1,20 @@
...
+    fun stay(): State
+
+    val isBlackjack: Boolean
+        get() = false
+
+    fun profit(
+        state: State,
+        betMoney: Int,
+    ): Double
+}
```
Wordbe: Usage of `State` 👍

Wordbe (changes requested):
Hello, Vito!

Great job creating and applying the concept of "State" in your Blackjack mission. 🚀
Through this, it’s a great opportunity to learn how polymorphism can be effectively used in object-oriented programming.

Overall, you've written impressive code.
There are no critical issues, but I’ve left some comments with suggestions—things that might be worth reconsidering or small improvements.
Please take a look and apply what you find helpful.

Thank you! 😊

src/main/kotlin/blackjack/model/Player.kt
VittorioDeMarzi: that make sense :)

src/main/kotlin/blackjack/model/BlackjackGame.kt
VittorioDeMarzi: I added a fun isRunning(): Boolean in the State :)

src/main/kotlin/blackjack/model/BlackjackGame.kt
VittorioDeMarzi: This is for the last two feedbacks. They let me think quite a lot and I came out with this approach.

Do you think it's idea hitting the first two cards in Participant like this?

```
Participant...{
  fun hitTwoCards(cardDeck: CardDeck) {
    while (state.isFirstTurn()) state = state.draw(cardDeck.drawCard())
  }
}
``` 

and call it :
    
```
fun setUpTable() {
  players.forEach { it.hitTwoCards(cardDeck) }
  dealer.hitTwoCards(cardDeck)
}
```

src/main/kotlin/blackjack/model/Card.kt
VittorioDeMarzi: You're right. Now that I've created an object pool and I'm no longer intending to create unique card instances on the fly, it doesn't make sense to keep Card as a data class anymore 😜

src/main/kotlin/blackjack/model/Dealer.kt
VittorioDeMarzi: It happened that I even moved the stay and draw to the Participant abstract class and it came out I guess a nicer and cleaner structure :)

```
class Dealer() : Participant("Dealer", FirstTurn(Hand(emptyList()))) {
  fun playTurn(deck: CardDeck) {
    while (state.isRunning() && state.hand.calculatePoints() < SEVENTEEN) {
      draw(deck.drawCard())
    }
    stay()
  }
}
```

let me please know what you think 😊

VittorioDeMarzi:
Thank you so much for your thoughtful review and valuable feedback! 🙏  
I’ve taken the time to go through each comment, and they really made me reflect on my code structure, especially regarding encapsulation and polymorphism.

I applied all your suggestions:
- ✅ Removed the unnecessary default constructor
- ✅ Refactored `Card` to no longer be a `data class` (since I now use an object pool)
- ✅ Moved `draw()` and `stay()` into the `Participant` class to avoid external state manipulation
- ✅ Replaced magic numbers with named constants
- ✅ Replaced `state is Running` with polymorphic checks like `state.isRunning()` to keep things cleaner

I also left a couple of questions open, like:  
> 💭 What do you think about the `hitTwoCards(cardDeck)` approach inside `Participant`?

If you still have suggestions for improvement, I’m totally open to them.  
I’d love to refine the code further if needed 😊

Thanks again for your time and for helping me grow through this review!

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,22 @@
...
+
+abstract class Participant(
+    val name: String,
+    var state: State,
+) {
+    val points get() = state.hand.calculatePoints()
+
+    fun draw(card: Card) {
+        state = state.draw(card)
+    }
```
Wordbe, line 14: 👍

src/main/kotlin/blackjack/model/BlackjackGame.kt
Wordbe: From the participant’s perspective, the initial state isn’t really about hitting twice — it’s more like simply receiving two cards.
So a name like `dealInitialCards` would be semantically clearer! 

In the same vein, instead of `while (FirstTurn())`, using something like `repeat(2)` makes the intent more explicit and easier to understand.
What do you think? 😊

src/main/kotlin/blackjack/model/Dealer.kt
Wordbe: It looks nicely cohesive—great job! 👍

src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,16 @@
+package blackjack.model
+
+import blackjack.states.FirstTurn
+
+class Dealer() : Participant("Dealer", FirstTurn(Hand(emptyList()))) {
```
Wordbe: You can probably remove the default constructor from `Dealer`, and it seems like `FirstTurn()` could be used to simplify the expression as well!

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -1,21 +1,7 @@
...
-        get() = numberInHand() == 2 && calculatePoints() == 21
-    val isBust
-        get() = calculatePoints() > 21
-
-    fun addCard(card: Card) = hand.addCard(card)
-
-    fun numberInHand() = hand.numberOfCards()
-
-    fun calculatePoints() = hand.calculatePoints()
+class Player(name: String) : Participant(name, FirstTurn(Hand(emptyList()))) {
```
Wordbe: Same here!

src/main/kotlin/blackjack/model/CardDeck.kt
```diff
@@ -1,24 +1,21 @@
 package blackjack.model
 
-data class CardDeck(private val cards: Cards) {
+data class CardDeck(private val hold: MutableList<Card>) {
     constructor() : this(initPokerCards())
```
Wordbe: Try to express the two constructors in CardDeck using a single one. 🙂
(Hint: with a default value)

src/main/kotlin/blackjack/model/CardDeck.kt
```diff
@@ -1,24 +1,21 @@
...
-        private fun list(): List<Card> {
-            return Suit.entries.flatMap { suit -> Rank.entries.map { rank -> Card(suit, rank) } }
-        }
-    }
+    val size
+        get() = cards.size
 
-    fun numberOfCards() = cards.cards.size
+    val cards
+        get() = hold.toList()
```
Wordbe: It might be easier to read if the type is explicitly specified!

src/main/kotlin/blackjack/model/CardDeck.kt
```diff
@@ -1,24 +1,21 @@
...
 
-    fun getCards() = cards.cards.toList()
+    fun drawCard(): Card {
+        val card = hold.first()
+        hold.remove(card)
+        return card
+    }
```
Wordbe: Try using `removeFirst()` once.
(Kotlin provides many utility functions like this. Try thinking if there's a more convenient function available.)
And it would be good to have some logic to handle the case when there are no cards left. 🙂

src/main/kotlin/blackjack/controller/BlackJackController.kt
```diff
@@ -1,27 +1,40 @@
...
+            players.forEach {
+                val bettingAmount = InputView.bettingAmountInput(it)
+                it.bettingAmount = bettingAmount
+            }
 
-            OutputView.printFinalResults(listOf(dealer) + players)
-            val statistics = Statistics(dealer, players)
-            OutputView.printStatistics(statistics)
+            val blackjackGame =
+                BlackjackGame(
```
Wordbe, line 22: A suggestion for the next step up (optional)
Try making the game run without needing a Controller or View.
In other words, aim to test `blackjackGame.play()` directly.

Wordbe (changes requested):
Hello Vito,

Overall, the code you wrote has a solid structure, clear responsibilities among objects, resulting in good cohesion and readability.
I really enjoyed reviewing it.

The review comments I left include minor things that could be fixed, as well as some points worth reconsidering further.
Your code is already great enough to merge, but I wanted to offer a bit more detailed feedback.
Please take a look and apply any parts you find feasible.

Thank you!

src/main/kotlin/blackjack/model/Player.kt
VittorioDeMarzi: I could easily cancel FirstTurn(Hand(emptyList())) in both of them since they are already initialized in Participant, FirstHand and Hand :)

src/main/kotlin/blackjack/model/CardDeck.kt
VittorioDeMarzi: Thanks for this feedback.
It really made me think. I think I found a nice kotlin idiomatic way to solve this. 
I added a control if the deck is empty before draw the new card. 🙂

src/main/kotlin/blackjack/controller/BlackJackController.kt
VittorioDeMarzi: Thanks for this advise. I will try for sure on my local to refactor it following your advice 😊

VittorioDeMarzi:
Thank you so much for this amazing and thoughtful review 🙏
Your feedback really helped me think deeper about Kotlin idioms and how to better manage state transitions and responsibilities across objects.
It definitely gave me new insight into writing more expressive and maintainable Kotlin code!

I’ve applied all your suggestions with care, and I can feel they pushed me one step forward in terms of writing more idiomatic, testable, and clean architecture.

As for further improvements, I completely agree with your suggestion to remove dependencies from the View and Controller. I’ll definitely explore refactoring BlackjackGame to make it runnable independently, so I can test play() in isolation.

I’m also considering adding an initial configuration to allow the user to choose how many decks to use, which could make the game setup more flexible.

Thanks again for sharing your time and knowledge, this was a huge learning moment for me! 😊

Wordbe (approved):
Hello Vito!

Thank you for the excellent code.
Even though you had already implemented the Blackjack mission very well, I ended up writing quite a long review.
Now that the mission is complete, I’m planning to merge it into the main code.
I really appreciate the time and effort you put into this.

I hope this experience becomes a great stepping stone for your journey ahead. Thank you. 🙏

PR #30, opened 2025-07-20
[Blackjack - Step2] Submit miooo0o's mission

miooo0o:
### Betting System Implementation:
- Add PlayerFactory to handle player creation with betting amounts
- Implement input validation for player names and bet amounts
- Add Validator class to ensure proper name and amount validation
- Support reading player bet amounts through InputView

### Payout and Result System:
- Add PayoutMultiplier property to Outcome enum for different win conditions
- Implement PlayerResult class with finalAmount calculation for individual player payouts
- Implement DealerEarning class to track dealer's total earnings
- Integrate PlayerResult and DealerEarning into comprehensive GameResult system

### Game Logic Improvements:
- Fix Blackjack detection logic to properly identify 21 with initial two cards
- Add dealer hit/stand logic with proper state transitions
- Implement edge case handling for various game outcomes
- Separate val/var usage to prevent state mutation issues

### Testing and Code Quality:
- Move test-specific classes (DummyDeck, DummyPlayerFactory) to test package
- Improve naming and readability 
- Collect view logic from Card and Suit classes into OutputView
- Reduce method visibility to internal where appropriate
- Add comprehensive test coverage for gameplay scenarios and edge cases

---

## Reflection

This is the first commit of step2, and I learned a lot about why forcing inheritance like this is problematic. What you mentioned about those 5 issues turned into an endless cycle when I tried to fix just the errors discovered in the process.

The most terrible part was the `state (var/val)` configuration being tangled up due to the `receive` method. Fixing one thing would break another, so I spent the entire day writing tests, writing more tests, writing even more tests, fixing, and repeating. Honestly, I'm still very anxious about it. Even though all tests pass, I'm worried.

Before moving to step2, I had prepared a revert version that only fixed what was reviewed. I wonder if I should have said I'd go with that version instead, given how painful this was. However, thanks to this struggle, I'm now more determined to avoid making the structure complicated!

### What I Regret
- It took so long to fix errors and the complex relationships made it hard to catch up once I lost focus, so I couldn't update the README in time, which is disappointing.
- I thought it would be better to separate `val playerResults: List<PlayerResult> = getAllPlayerResults()` into its own class (since there's evaluation logic, it seemed better to extract it), but due to time constraints, I decided to send the PR first. I'm curious about your thoughts on this.

### What I'd Like Feedback On
- I tried to make it at least a bit more readable without changing the structure too much. Do you think it improved a little?
- I did untangle the `state (var/val)` and `receive` (previously, receive directly mutated state), but I felt there were too many places tracking state changes with `get()`. Even when running the debugger, it was hard for me to follow... For values that "need to continuously change or be updated," what would be a good way to organize them?
- I'm sure there are probably more errors somewhere. During this change, I barely thought about questions like "why is this class a class?" I was just grateful when things worked and was afraid to change more... I learned again that not writing tests properly leads to pain. If you notice any areas I should cover better, please let me know.


src/main/kotlin/blackjack/factory/PlayerFactory.kt
```diff
@@ -0,0 +1,20 @@
+package blackjack.factory
+
+import blackjack.model.participant.Player
+import blackjack.validator.Validator
+
+interface BasePlayerFactory {
+    fun createPlayers(): List<Player>
+}
```
pkeugine:
```suggestion
fun interface BasePlayerFactory {
    fun createPlayers(): List<Player>
}
```
We can make this a functional interface if we have only one method in it.
It would be great to check out what functional interface is :)

src/main/kotlin/blackjack/model/holder/CardHolder.kt
```diff
@@ -9,27 +9,26 @@ abstract class CardHolder {
...
+    fun draw(count: Int): List<Card> {
+        if (currentCards.isEmpty() || currentCards.size < count) {
             onDrawFailed()
         }
+        val cards = currentCards.take(count)
+        repeat(count) { currentCards.removeFirst() }
+        return cards
+    }
+
+    /** fallback when cannot draw */
```
pkeugine, line 26: Since `onDrawFailed` seems like it describes itself pretty well, I don't think we need this comment here. Let's keep that in mind when working on other projects.

You might have noticed I put very close attention to comments in between logics.
Although it might seem like they're not that big of a deal, I prefer adding comments when I feel they are **extremely necessary**. If not, the whole team starts to add in comments and they become part of maintaining. That leads to deprecated comments, or comments that are only understandable to certain people. This could be a real pain, because it's not always clear if I can remove that comment or not, resulting in confusing comments staying there forever.

I do understand that `fallback when cannot draw` seems like it won't confuse us in the future. But I thought I would explain why I pay close attention to comments since I've left reviews about it in the previous PR as well.

src/main/kotlin/blackjack/model/participant/Dealer.kt
```diff
@@ -1,33 +1,42 @@
...
-            State.HIT
-        } else {
-            State.STAY
-        }
+    override fun isBlackjack() = state == State.BLACKJACK && score == BLACKJACK_SCORE
+
+    internal fun calculateState(): State {
+        if (isFirstRound()) return initialStateOrBlackjack()
+        return evaluateState()
     }
```
pkeugine: I've noticed you've been using internal access level for some methods.
I'm guessing this is due to test availability for private methods.

However, some of these internal methods aren't tested, and are internal but not used in anywhere else.
I would keep them private, and focus on testing public methods that use these private methods.
If we make good tests and cover the edge cases for public method tests, I'm sure we'll be able to cover private methods as well.

src/main/kotlin/blackjack/model/holder/CardHolder.kt
```diff
@@ -9,27 +9,26 @@ abstract class CardHolder {
...
     /**
      * @throws IllegalStateException if try failed and should throw error
      *
      */
-    fun draw(): Card {
-        return if (currentCards.isNotEmpty()) {
-            currentCards.removeFirst()
-        } else {
+    fun draw(count: Int): List<Card> {
+        if (currentCards.isEmpty() || currentCards.size < count) {
```
pkeugine, line 18: Nice!! 💯 💯 💯

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -1,62 +1,78 @@
...
+            Suit.DIAMOND -> "♦"
+            Suit.CLUB -> "♣"
+            Suit.SPADE -> "♠"
+        }
 
-        playersResults.forEach { (player, outcome) ->
-            println("${player.name}: ${outcome.name}")
+    internal fun showGameSummary(gameResult: GameResult) {
+        println("\n${Message.FINAL_RESULTS_TITLE}")
+        println("Dealer: ${gameResult.dealerEarning.amount}")
```
pkeugine: We can use Kotlin's [multi-line string feature](https://kotlinlang.org/docs/java-to-kotlin-idioms-strings.html#use-multiline-strings) when printing multiple lines!
`println` is actually quite expensive when it comes to using computation resources, so we might as well decrease the amount of times calling it. It might be fun to search and learn why :)

src/test/kotlin/blackjack/controller/GameManagerTest.kt
```diff
@@ -15,29 +17,29 @@ import org.junit.jupiter.api.Test
...
     @Test
     fun setUp() {
-        val participants = Participants.from(listOf("mina", "guri"))
+        val participants = Participants(DummyPlayerFactory(2).players)
         GameManager(participants).setUp()
 
-        OutputView.showCards(participants.getDealer())
-        OutputView.showAllPayersCards(participants)
+        OutputView.showCards(participants.dealer)
+        OutputView.showAllPayersCards(participants.players)
```
pkeugine: Although I didn't leave this review in the previous PR, it would be great to have at least one assertion in tests instead of having UI result. This is because we don't actually see the whole test result logs whenever we run the tests. That means even if the output is wrong, the test will still pass and we won't notice it. In this case, I don't think this test is necessary, with reasons I've written in the previous comment.

src/test/kotlin/blackjack/model/participant/DealerTest.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    @Test
+    fun `dealer busts when score exceeds 21`() {
+        val dealer = Dealer()
+        dealer.receive(TWO_CARDS_SUM_16)
+        dealer.receive(listOf(TestFixture.Card.DIAMOND_JACK))
+
+        println("cards: ${dealer.cards}")
+        println("score: ${dealer.score}")
+        println("state: ${dealer.state}")
```
pkeugine: Tests don't need println in their final form. They can be used when we try to debug a test, but let's keep them out of our main branch.

src/test/kotlin/blackjack/model/result/PlayerResultTest.kt
```diff
@@ -0,0 +1,39 @@
...
+        assertThat(playerResult.finalAmount).isEqualTo(-10000)
+    }
+
+    @Test
+    fun `final amount is 20000 if player bet 10000 and Win`() {
+        val player = DummyPlayerFactory(1).players[0]
+        val playerResult = PlayerResult(player, Outcome.WIN)
+
+        assertThat(playerResult.finalAmount).isEqualTo(20000)
+    }
```
pkeugine, line 22: I can see that you've put so much effort in testing 👍 

**PR message: Honestly, I'm still very anxious about it. Even though all tests pass, I'm worried.**
Don't be. You won't last so long as a developer with that anxiety in you. Just let it go, and fix when somebody finds a bug 😄

README.md
```diff
@@ -2,66 +2,25 @@
...
-- Card values follow standard Blackjack rules:
-- Number cards are counted by their face value.
-- Face cards (King, Queen, Jack) are each worth 10.
-- Aces can be worth either 1 or 11.
-- Each player starts with two cards.
-- Players may draw additional cards as long as their total remains 21 or less.
-- The dealer must draw a card if their total is 16 or less, and must stand on 17 or more.
-- If the dealer busts (goes over 21), all remaining players automatically win.
-- After the game ends, display the result (win/loss) for each player.
+- [x] Each player must place a bet at the start of the game.
```
pkeugine, line 5: **PR message : thanks to this struggle, I'm now more determined to avoid making the structure complicated!**

**PR message 2 : I tried to make it at least a bit more readable without changing the structure too much. Do you think it improved a little?**

Yes, the project looks much better than before, and you've done a great job refactoring it. I wouldn't mind making it more simple though, like I said in the previous comments. No worries about testing private functions, no UI related testing when making unit tests, no comments when the code describes itself pretty well, etc. Personally, I like to think as : `Simple is better most of the times. Add in stuff when people say they need it` 😄

README.md
```diff
@@ -2,66 +2,25 @@
 
 ## Requirements
```
pkeugine, line 3: **PR message : I thought it would be better to separate `val playerResults: List<PlayerResult> = getAllPlayerResults()` into its own class (since there's evaluation logic, it seemed better to extract it), but due to time constraints, I decided to send the PR first. I'm curious about your thoughts on this.**

Yes, I do get what you mean when you say there could be a class dedicated to `List<PlayerResult>`, but the current logic seems fine to me. I'm fine with keeping it as is. Though I think `DealerEarning` is the class that you're looking for; it just has a name related to `Dealer`. 

(I'm not saying it's wrong, so don't stress too much about refactoring this)

pkeugine, line 3: **PR message : For values that "need to continuously change or be updated," what would be a good way to organize them?**

For values that need to change over time, using var is typically appropriate 😅.
Alternatively, if you want to keep things immutable, consider returning a new object with the updated values instead of mutating the existing one, a common pattern in functional programming.

Most of the times in web backend development, we use our database 😎

pkeugine, line 3: **PR message : During this change, I barely thought about questions like "why is this class a class?" I was just grateful when things worked and was afraid to change more... I learned again that not writing tests properly leads to pain. If you notice any areas I should cover better, please let me know.**

I'm continuously noticing you're thinking a bit too deep about this.
Let's try to keep things (including logic) simple, not worry too much about it, and let the problem occur... cuz we're never free from bugs 🙂

Hey, even the inventors of C language, Linux, and all these popular software created bugs and legacy codes that others couldn't deal with. This happens to everyone, and nobody is free from it. So we might as well be cool with it.

pkeugine (changes requested):
Hello Mina,
Well done keeping this mission going even when you're busy with the next mission.
I tried to keep the review simple, since it is a fact that `we're currently working on the next mission`. However, I thought you might want to look at the comments and refactor a bit, since I know you well (... I think).

But I want to give you an additional goal:
**Keep it under 5 commits**.

Once you're done, give me an additional review request, and I'll merge this PR 👍 

Happy Hacking, and remember, **keep it simple**. 😎

src/test/kotlin/blackjack/model/result/PlayerResultTest.kt
miooo0o: I wanted to make sure I had at least the basic tests covered before PR.

I realized during the Lotto mission that if I write loose tests, I can go on for a long time thinking everything’s fine when it’s actually not. This time I was writing and editing tests again and I actually found a lot of tests that made no sense haha. but... sure, if something breaks, I can just fix it!

src/test/kotlin/blackjack/model/participant/DealerTest.kt
miooo0o: Oh, this is the part I forgot to delete. I'll delete!

src/test/kotlin/blackjack/controller/GameManagerTest.kt
miooo0o: That's right. I completely forgot this test during refactor.
I'll have to spend some more time cleaning up the code. Thank you, I'll apply it!

src/main/kotlin/blackjack/view/OutputView.kt
miooo0o: Oh wow, I never thought `println()` could be expensive. that’s really surprising! I definitely want to look into why that is. It’s super interesting, and I’d love to try applying it.

README.md
miooo0o: It kind of feels like if I don't deal with it now, it'll come back to haunt me later like karma. still but slowly I let go.
This time, not overthinking actually made things a bit easier. Just letting it be like, “Okay, this works, note and move on” felt really good.

I guess learning to let go of the pressure is part of growing :confetti_ball:

miooo0o: This is my first time doing something where there is a distinction between `var` and `val`, so I felt like I shouldn't use `var`. I don't know why but :crying_cat_face: 

If you use a database in the backend, there must be a convention, right? You mean I have less to worry about? It would be great.

miooo0o: Ah, I’m glad to hear that!  I’ll have to extract the dealer logic separately. I’ll go ahead and do that.

Coding itself doesn’t stress me out! If anything, all the overthinking and constant questioning is just a leftover occupational habit from the past. Honestly, I kind of enjoy that kind of stress :sparkles: 

I wanted to hit my goals (like sticking to the basics and improving test quality), so I was a bit worried that I didn’t quite get there.  But aside from that, I genuinely enjoyed it.

src/main/kotlin/blackjack/model/participant/Dealer.kt
miooo0o: I tried my best to find and replace as many as I could!

src/main/kotlin/blackjack/model/holder/CardHolder.kt
miooo0o: I see that proper documentation really makes a difference. I used to just write documents from time to time, but intentional documentation is definitely something else! 
As always, keep it simple and understandable through code only.

Thanks for the advice.

src/main/kotlin/blackjack/factory/PlayerFactory.kt
miooo0o: As I understand it so far, a `fun interface` allows us to pass a lambda.
I’m not exactly sure how I would adapt that in the current version. 
But if I ever need a new kind of test, I’ll be able to experiment with a lambda instead of creating a new class. right?

For now, I’ve just converted the `interface` to a `fun interface`.
Please have a look, and let me know if you’d recommend writing a test to try it out!

README.md
miooo0o: In this set of commits, I removed all the test output that was previously being printed.It definitely looks much cleaner now. Yes, simple is better most of the time. happier, simpler and cleaner!

miooo0o:
### **Step 2-2 Review Feedback**

Hi PK, 
Thank you for helping me see where I can improve.I learn a lot from you!

#### The following changes are included in this set of commits:
- Limited to 5 commits
- Changed visibility of untested methods to `private`
- Extracted `DealerEarning` init from `GameResult`
- Removed `println()` and` OutputView` calls from tests
- Cleaned up unnecessary test
- Deleted redundant comment in `CardHolder`
- Changed BasePlayerFactory to `fun interface`

---

- This was my first time using git squash. I used to commit everything at once, so I focused only on splitting commits. I hadn't really thought about how to merge them back. It feels good to learn a new tool. 
- Extracting `DealerEarning` from `GameResult` was surprisingly simple. lovely.

As I mentioned above, if there's anything you'd like me to try out in Step 2, feel free to let me know.
I can work on it little by little over the week/end!

Thank you again and have a great day.

src/main/kotlin/blackjack/factory/PlayerFactory.kt
pkeugine: Yes, you can use it in lambda 👍
That's it! You don't need to do anything.

README.md
pkeugine: I think we'll cover this as we progress through the next mission :)

pkeugine (approved):
Hello Mina,
Nice work refactoring the project in 5 commits 👍

Like I said before, I think it's time we finish this blackjack mission. I hope you do well on your next missions, and if there are any questions or thought, feel free to DM me anytime.

Happy Hacking! 🌟🌟🌟

PR #32, opened 2025-07-22
[Blackjack - Step2] Submit san-ghun's mission

san-ghun:
# Blackjack Step 2

## Overview

This PR is to implement requirements of the Blackjack project step2. The project is a simulation of card game Blackjack in terminal text input and output. The step2 requirement is to add betting feature in the version of Blackjack. In this game, player(s) bet certain amount, the player or dealer with a total closest to 21 without going over wins, and finally display earnings for player and dealer based on player's betting and game result.

### `Result` has new enum `BLACKJACK` 

- It is to mark the result of player or dealer who has blackjack, for player to receive 1.5x of their earning.

### `Player` can bet before the initial draw of cards

- `Playable` has a property named `bet` with integer value to store betting amount.
- `Playable` has a method `isBlackjack()` to figure out the `Playable` is blackjack or not.
- In 'InputView', `readPlayerBettingAmount()` read user input of player's betting amount.
- `Player` has a method `placeBets()` to add betting amount to on its `bet` property.

### Output game result with earnings of all `Playable`s

- `Stats` calculate pay-outs for players and dealer based on the conditions;
    - All betting values are added on a variable `pot` at during the `payOutPotToEarnings()` method.
    - Check the dealer has blackjack or not.
    - Check the player has blackjack or not.
    - All players get pay-outs from `pot`, then dealer get reminder from `pot`.
- 'OutputView' display the final result and earnings with `displayEarnings()`.

## Additional Notes

- Try to simplify the interaction between functions to improve readability.
- Try to remove states and replace those into methods in class.

In this project, coding logic for calculating the earnings for `Playable`s was difficult, especially to separate the logic into several functions to meet the Programming Requirements of this project. Also tried to apply reviewed comments from previous step1, for example, replacing states into methods.

Thank you!

### DIFF
--- a/src/main/kotlin/blackjack/controller/GameMaster.kt
+++ b/src/main/kotlin/blackjack/controller/GameMaster.kt
@@ -16,8 +16,8 @@ object GameMaster {
         askPlayersToHit(players)
         drawDealerCards(dealer)
         OutputView.displayFinalState(players, dealer)
-        val winStatistics = calculateStatistics(players, dealer)
-        OutputView.displayFinalResults(winStatistics)
+        val winStatistics = Stats(players, dealer)
+        OutputView.displayEarnings(winStatistics)
     }
 
     private fun initPlayers(): List<Player> {
@@ -30,27 +30,22 @@ object GameMaster {
     }
 
     private fun askPlayersToHit(players: List<Player>) {
-        val deck = PlayingCard.deck
         players.forEach { player ->
-            askPlayerHit(player) { deck.giveCard() }
+            askPlayerHit(player)
         }
     }
 
-    private fun askPlayerHit(
-        player: Player,
-        receiveCard: () -> PlayingCard,
-    ) {
+    private fun askPlayerHit(player: Player) {
         var isFirst = true
 
         while (!player.isBust()) {
-            isFirst = drawOrNot(player, receiveCard, isFirst)
+            isFirst = drawOrNot(player, isFirst)
             if (isFirst) break
         }
     }
 
     private fun drawOrNot(
         player: Player,
-        receiveCard: () -> PlayingCard,
         isFirst: Boolean,
     ): Boolean {
         val wantsCard =
@@ -63,15 +58,13 @@ object GameMaster {
             return true
         }
 
-        drawAndDisplayAndReturnFalse(player, receiveCard)
+        drawAndDisplayAndReturnFalse(player)
         return false
     }
 
-    private fun drawAndDisplayAndReturnFalse(
-        player: Player,
-        receiveCard: () -> PlayingCard,
-    ): Boolean {
-        player.drawCard(receiveCard())
+    private fun drawAndDisplayAndReturnFalse(player: Player): Boolean {
+        val deck = PlayingCard.deck
+        player.drawCard(deck.giveCard())
         OutputView.displayCurrentHand(player)
         return false
     }
@@ -82,14 +75,5 @@ object GameMaster {
             dealer.drawCard(deck.giveCard())
             OutputView.displayDealerDrawsCard()
         }
-    }
-
-    private fun calculateStatistics(
-        players: List<Player>,
-        dealer: Dealer,
-    ): Stats {
-        val winStatistics = Stats(players, dealer)
-        winStatistics.updateDealerStats()
-        return winStatistics
     }
 }
--- a/src/main/kotlin/blackjack/controller/GameMaster.kt
+++ b/src/main/kotlin/blackjack/controller/GameMaster.kt
@@ -10,7 +10,7 @@ import blackjack.view.OutputView
 object GameMaster {
     fun run() {
         PlayingCard.deck.shuffle()
-        val dealer = Dealer()
+        val dealer = Dealer().initHand()
         val players = initPlayers()
         OutputView.displayInitialState(players, dealer)
         askPlayersToHit(players)
@@ -25,7 +25,7 @@ object GameMaster {
         val bets = names.map { InputView.retryable { InputView.readPlayerBettingAmount(it) } }
         val players =
             names.zip(bets)
-                .map { (name, amount) -> Player(name).placeBets { amount } }
+                .map { (name, amount) -> Player(name).placeBets(amount).initHand() }
         return players.toList()
     }
 
--- a/src/main/kotlin/blackjack/controller/GameMaster.kt
+++ b/src/main/kotlin/blackjack/controller/GameMaster.kt
@@ -10,7 +10,7 @@ import blackjack.view.OutputView
 object GameMaster {
     fun run() {
         PlayingCard.deck.shuffle()
-        val dealer = Dealer().initHand()
+        val dealer = Dealer()
         val players = initPlayers()
         OutputView.displayInitialState(players, dealer)
         askPlayersToHit(players)
@@ -25,7 +25,7 @@ object GameMaster {
         val bets = names.map { InputView.retryable { InputView.readPlayerBettingAmount(it) } }
         val players =
             names.zip(bets)
-                .map { (name, bet) -> Player(name).placeBets(bet).initDraw() }
+                .map { (name, bet) -> Player(name, bet) }
         return players.toList()
     }
 
@@ -38,7 +38,7 @@ object GameMaster {
     private fun askPlayerHit(player: Player) {
         var isFirst = true
 
-        while (!player.isBust()) {
+        while (!player.hand.isBust) {
             isFirst = drawOrNot(player, isFirst)
             if (isFirst) break
         }
--- a/src/main/kotlin/blackjack/controller/GameMaster.kt
+++ b/src/main/kotlin/blackjack/controller/GameMaster.kt
@@ -28,12 +28,13 @@ object GameMaster {
     }
 
     private fun initPlayers(): List<Player> {
-        val players = mutableListOf<Player>()
-        val names = InputView.retryable { InputView.readPlayerNames() }
-        names.forEach { name ->
-            val player = Player(name, Hand().initCards())
-            players.add(player)
-        }
+        val players = InputView.retryable { InputView.readPlayerNames() }
+            .map { Player(it, Hand().initCards()) }
+            .map { player ->
+                player.placeBets {
+                    InputView.retryable { InputView.readPlayerBettingAmount(player.name) }
+                }
+            }
         return players.toList()
     }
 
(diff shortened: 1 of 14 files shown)


src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -14,6 +15,11 @@ data class Dealer(override val name: String = "Dealer") : Playable {
         return hand.calculateHand()
     }
 
+    fun initHand(): Dealer {
+        hand = Hand().initCards()
+        return this
+    }
```
Wordbe: If you store the hand directly during Dealer initialization, I don’t think `initHand()` is necessary. Just calling the constructor should be enough. What do you think? ☺️

src/main/kotlin/blackjack/controller/GameMaster.kt
```diff
@@ -11,54 +10,42 @@ import blackjack.view.OutputView
...
-        val players = mutableListOf<Player>()
         val names = InputView.retryable { InputView.readPlayerNames() }
-        names.forEach { name ->
-            val player = Player(name, Hand().initCards())
-            players.add(player)
-        }
+        val bets = names.map { InputView.retryable { InputView.readPlayerBettingAmount(it) } }
+        val players =
+            names.zip(bets)
+                .map { (name, amount) -> Player(name).placeBets(amount).initHand() }
```
Wordbe: Using `zip` is a great choice! 😊 That said, it might be better to stick to just one term—either `bet` or `amount`—for consistency. What do you think?

src/main/kotlin/blackjack/model/Playable.kt
```diff
@@ -17,7 +18,14 @@ interface Playable {
         return calculateHand() > BUST_LIMIT
     }
 
+    fun isBlackjack(): Boolean {
+        val hasTwoCards = hand.cards.size == 2
```
Wordbe: I think `hasTwoCards` could be a responsibility of the `Hand` class.
Also,try to create a test of `Hand` and eliminate the magic number 😊

src/main/kotlin/blackjack/model/Playable.kt
```diff
@@ -17,7 +18,14 @@ interface Playable {
...
+        val hasTwentyOne = calculateHand() == BUST_LIMIT
```
Wordbe: Now that I look at it, `calculateHand()` could be implemented directly in the interface since both implementations use the same logic 😊
With that in mind, I think `hasTwentyOne` could also be a responsibility of the `Hand` class.

src/main/kotlin/blackjack/model/Player.kt
```diff
@@ -12,4 +14,14 @@ data class Player(override val name: String, override var hand: Hand = Hand()) :
...
+
+    fun initHand(): Player {
+        hand = Hand().initCards()
+        return this
+    }
+
+    fun placeBets(amount: Int): Player {
+        bet += amount
+        return this
+    }
```
Wordbe: Same here as with the Dealer—looks like the constructor can handle it just fine! 😊
Also, try adding a validation check inside the constructor.

src/main/kotlin/blackjack/model/Result.kt
```diff
@@ -4,5 +4,11 @@ enum class Result() {
     LOSE,
     WIN,
     TIE,
+    BLACKJACK,
     NONE,
+    ;
+
+    companion object {
+        const val BLACKJACK_BONUS = 1.5
```
Wordbe: Specifying the type makes it easier for others to read the code 🙂

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -39,25 +38,14 @@ object OutputView {
...
-            Result.LOSE -> "Lose"
-            Result.WIN -> "Win"
-            Result.TIE -> "Tie"
-            else -> "Error"
+    fun displayEarnings(winStatistics: Stats) {
+        val players = winStatistics.players
+        val dealer = winStatistics.dealer
+        val earningsMap = winStatistics.payOutPotToEarnings()
+        println("\n## Final Earnings")
+        println("Dealer: ${earningsMap[dealer]}")
```
Wordbe: How about calling `println` once to check? 🙂

src/main/kotlin/blackjack/model/Stats.kt
```diff
@@ -1,20 +1,6 @@
 package blackjack.model
 
 class Stats(val players: List<Player>, val dealer: Dealer) {
```
Wordbe, line 3: The game result calculation and payout logic are mixed together—it would be better to separate them 😊
Also, can you try simplifying the code a bit more?
For example, instead of subtracting each player's `payoutImpactOnPot` from the pot, you could try something like:

```kotlin
val dealerEarnings = totalBets - totalPlayerPayouts
```

This way, you can eliminate mutable variables.
Try making good use of functional programming too!

Wordbe (changes requested):
Hello San!

Thank you for understanding and implementing the additional requirements of the Blackjack mission so well.
I’ve left comments on some points that need further thought, so please review them and apply any changes where possible.
Also, please strengthen the tests for the parts you modified, if you can (Dealer, Playable).

Thank you! 😊

src/main/kotlin/blackjack/controller/GameMaster.kt
san-ghun: I agree, good point! Thank you!

src/main/kotlin/blackjack/model/Playable.kt
san-ghun: Yes, it seems better fit with `Hand`!

src/main/kotlin/blackjack/model/Playable.kt
san-ghun: Agree! I move this to `Hand`.

src/main/kotlin/blackjack/model/Result.kt
san-ghun: Yes, it would be better with type notation. I will keep that in mind! Thanks!

src/main/kotlin/blackjack/view/OutputView.kt
san-ghun: What you want to say is, make the two lines of `println` into one line, like below?
So, to reduce the amount of calling `println` function?
Did I understood correct?

```kotlin
println("\n## Final Earnings\nDealer: ${earningsMap[dealer]}")
```

src/main/kotlin/blackjack/model/Dealer.kt
san-ghun: Yeah, your right. 
I was wondering how I could write tests that could inject a set of cards and control conditions to keep the outcome of the `Playable` predictable, without having to control the injection of cards into the `Playable` using `initHand()`.

I come up with this solutions, :

1. Let's inject `Hand` into `Playable`.

```kotlin
data class Dealer(
    override val name: String = "Dealer",
    override val hand: Hand = Hand(),
) : Playable {
//...
}
```

2. Let `Hand` to init with condition.

```kotlin
class Hand(var cards: List<PlayingCard> = emptyList()) {
    val isBust: Boolean get() = calculateHand() > BUST_LIMIT
    val isBlackjack: Boolean get() =
        (cards.size == INIT_CARDS_SIZE) && (calculateHand() == BUST_LIMIT)

    init {
        if (cards.isEmpty()) {
            val deque = ArrayDeque(cards)
            repeat(INIT_CARDS_SIZE) {
                deque.addLast(PlayingCard.deck.giveCard())
            }
            cards = deque.toList()
        }
    }
//...
}
```

src/main/kotlin/blackjack/model/Player.kt
san-ghun: Yes, I apply same way to init `Player` with constructor like did on `Dealer`.
However, I didn't get what you mean that, try adding a validation check inside the constructor, for which validation you mean?

src/main/kotlin/blackjack/model/Stats.kt
san-ghun: Yes, I can see some chance to simplify the code. I will aggregate both logic into one and make it simple!

san-ghun:
Hello Suno!

Wish you the best of this summer.
Always thanks for your review with valuable suggestions.
I tried to apply your reviews.

## Main Changes

### Move actions and properties from `Playable` to `Hand`

- transfer calculate-able properties `isBust` and `isBlackjack` to check the status of 'hand' from `Playable` to `Hand`.
- remove `calculateHand()` from `Playable` and call to `Hand` directly to calculate the hand.
- remove `initHand()` from `Player` and `Dealer` to `Hand` by applying `init{}` with condition to initiate the `Hand`.
- remove `placeBets()` from `Player` and use constructor to get `bet` amount, since the betting is done at once at the beginning of the game.

### Aggregate and simplify calculation of payout and earning of `Playable`s

- aggregate and simplify the calculation of payout and earning of `Player` and `Dealer` by applying the functional programming approach.

## Additional Note

Through your comments and review, I tried to apply functional programming as much as possible. I realized that there are plenty way of using the functional programming approach to implement even a single logic. It was fascinating to have multiple choices and to find out more options by chaining other methods.

Also, I found that it is essential to have an Object-Oriented mindset like using constructor efficiently and take advantage of OOP way. I am still finding out and learning the OOP way through projects and also from reading articles about the Design Patterns. Those are still hard to understand and apply for me, however, I am getting more curious what will come next and how my code will become more magical like Jason, Brie and Justin, our coaches.

I really appreciate your patience for my request for review.
It took long time to submit this request for review and I apologize for taking this long time.
However, thanks to you patience I could have more time to consider how to use functional programming to to simplify complex logic and apply the constructor to initialize an instance in more OOP way.
Thanks for keep me motivated through your thoughtful review and comments.

src/main/kotlin/blackjack/model/Stats.kt
Wordbe: So great. Thank you!

src/main/kotlin/blackjack/model/Player.kt
Wordbe: I meant it would be good to validate whether `amount` is a positive number! 🙂

src/main/kotlin/blackjack/view/OutputView.kt
Wordbe: I was referring to using the `"""` syntax for string templates. Please take a look! 📄
https://kotlinlang.org/docs/strings.html#string-templates

Wordbe (approved):
Hello San!

You didn’t give up and completed the mission to the end. great job!
Thank you for carefully reading through the reviews and applying the feedback.

I hope what you’ve learned from this mission will be useful in future projects as well.
If you take some time to revisit the following keywords, it’ll help solidify your understanding:

* Proper use of inheritance and interfaces
* Assigning responsibilities to objects to increase cohesion
* Applying polymorphism

I'll go ahead and approve and merge this PR now.
It would be great if you could also reflect on the additional comments I left.

Thank you again for your dedication all the way through! 🙌

PR #33, opened 2025-07-27
[Blackjack Step 2] Submit Petra's Mission

pebencze:
Hello Dani! 

Hope you are doing well! Sorry for the delay, I got caught up in the new missions... 

I have implemented the state machine inside my Blackjack step 2 and I am curious what you think about it. 
Dealer and players can have 5 different states based on the game rules: Bust, BlackJack, Stay, Hit and Init. Drawing and profit calculation happen inside the states. They are divided into two groups defined by two interfaces: Running and Stopped. It was initially difficult to implement it, but by now it seems quite easy and rewarding.

What is your opinion about my Controller and general structure of the project? 

Best,

Petra


pebencze:
Conflicts resolved, sorry for the mess. Probably there is a better way to do it... The application is working now.

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -1,95 +1,73 @@
...
+import blackjack.model.cards.CardDeck
+import blackjack.model.table.Dealer
+import blackjack.model.table.Player
+import blackjack.model.table.Players
+import blackjack.view.ErrorMessage
 import blackjack.view.InputView
 import blackjack.view.OutputView
-import blackjack.view.OutputView.displayInitialCardsMessage
 
 class Controller {
```
da-nyee, line 14: image

Working well 👍

src/main/kotlin/blackjack/controller/Controller.kt
```diff
@@ -1,95 +1,73 @@
...
     }
 
-    private fun initializePlayers(): Players {
+    private fun setBettingAmount(player: Player) {
         repeat(MAX_TRIES) {
             try {
-                val playerNames = InputView.readNames()
-                val players = Players(playerNames.map { Player(it) })
-                return players
+                player.bet = Bet.of(InputView.readBettingAmount(player))
```
da-nyee, line 45: Wrapped `Bet` as well 👍

src/main/kotlin/blackjack/model/Bet.kt
```diff
@@ -0,0 +1,18 @@
+package blackjack.model
+
+@JvmInline
+value class Bet private constructor(val amount: Int) {
+    companion object {
+        const val MIN_BET = 5
+        const val MAX_BET = 100
```
da-nyee: They could be `private`, since they're only using in this class 👀

src/main/kotlin/blackjack/model/Bet.kt
```diff
@@ -0,0 +1,18 @@
+package blackjack.model
+
+@JvmInline
+value class Bet private constructor(val amount: Int) {
```
da-nyee, line 4: `@JvmInline` + `value class` 👍

src/main/kotlin/blackjack/model/Bet.kt
```diff
@@ -0,0 +1,18 @@
...
+    companion object {
+        const val MIN_BET = 5
+        const val MAX_BET = 100
+
+        fun of(amount: Int): Bet {
```
da-nyee, line 9: Static factory method 👍

src/main/kotlin/blackjack/model/states/State.kt
```diff
@@ -0,0 +1,18 @@
+package blackjack.model.states
+
+import blackjack.model.Bet
+import blackjack.model.cards.Card
+import blackjack.model.cards.HandCards
+
+interface State {
```
da-nyee, line 7: You clearly know how to use the State Pattern 👍

src/main/kotlin/blackjack/model/cards/HandCards.kt
```diff
@@ -1,14 +1,9 @@
...
-        get() = calculateTotal()
+class HandCards(val cards: List<Card> = listOf()) {
+    operator fun plus(card: Card): HandCards = HandCards(cards + card)
 
-    fun add(card: Card) {
-        cards.add(card)
-    }
-
-    private fun calculateTotal(): Int {
+    fun total(): Int {
```
da-nyee, line 6: You might consider writing unit tests for this method 👀

README.md
```diff
@@ -88,6 +152,8 @@ jason: Lose
 - [x] display results
 
 ## Considerations
```
da-nyee: > What is your opinion about my Controller and general structure of the project?

It seems clean and well-organized overall — including the Controller 👍

README.md
```diff
@@ -88,6 +152,8 @@ jason: Lose
...
-Q1: How should we name players inside the Players class? </br>
-Q2: Should we use multiple decks? </br>
-Q3: State in Controller (even if private) is bad practice?</br>
\ No newline at end of file
+- Q: Would it make sense to directly call draw twice in the init block when creating an Init object?
```
da-nyee: `Init` currently looks like:

```kotlin
class Init(override val hand: HandCards = HandCards()) : Running {
    init {
        if (hand.cards.size > 1) {
            throw IllegalStateException()
        }
    }
```

Do you mean you’re considering refactoring this part?
I personally think the current implementation looks fine as it is 👀

README.md
```diff
@@ -88,6 +152,8 @@ jason: Lose
...
+- Q: Why do I not pass the whole deck as parameter to draw()?
```
da-nyee: Yeah, you could be right — but I think it looks good enough for now 🤔

README.md
```diff
@@ -88,6 +152,8 @@ jason: Lose
...
+- Q: Is it not costly to always return new instances of State?
```
da-nyee: It might add a bit of overhead, but modern memory management is pretty strong, so I don't think it's a big deal.

README.md
```diff
@@ -88,6 +152,8 @@ jason: Lose
...
+- Q: Mutable handcards vs immutable handcards -> adding a new card always returns a new instance of HandCards
```
da-nyee: I'd recommend using immutable `HandCards`, since it's a practice project.
It's not a large-scale system, so returning a new object each time shouldn't be a big concern.

da-nyee (changes requested):
Hi Petra! 🙌
You’ve done awesome 💯

It looks solid — I think it's ready to wrap up the mission.
I’ll leave this as a request change for now, since I’ve left a few comments — please take a look.
If you have any other questions, feel free to leave a comment and request a review again.
If everything’s clear, just let me know — I’ll go ahead and merge the PR ☺️

README.md
pebencze: I think the problem would be then that I could not change the state from within the class... so it does not work. I just forgot to delete the question from my README :)

README.md
pebencze: How do you decide on these things if you are building a big project?

pebencze:
Hello Dani!

Thanks for your insightful and encouraging feedback! 
I have implemented the small suggested changes and some additional tests.

Let me know if you think it is ready to be merged. :)

Best,

Petra

README.md
da-nyee: @pebencze 

Well, when I work on a larger project, I often end up reusing an existing object instead of always creating a new one.
Ideally, we should wrap a new object every time — that's what we've learned — but in reality, it's not always that easy.
It's one of the trade-offs we sometimes have to make.

da-nyee (approved):
Hi Petra! 🙌

It's finally time to wrap up the mission — congratulations 🎉
I could clearly see how much you've grown through this one — great job 💯

I left one last comment [here](https://github.com/herotechcourse/kotlin-blackjack/pull/33#discussion_r2242709218), so feel free to take a look when you have time.
If you have any further questions, don't hesitate to reach out — I'll be happy to help ☺️

PR #34, opened 2025-08-26
[Blackjack Step 2] Submit knopgm's Mission

knopgm:
Hey Suno!

I'm very very sorry for this huge late PR!
I had to deal with problems at the begginning of the course and that had impacted my performance during the whole course. 
But I made it! 💪 😄 

For this PR I’ve structured the code using OOP and the State pattern to keep the game logic clean and extensible.
The game flow is managed by a BlackjackController, which handles player turns, dealer logic, and user input/output.
The state implementation was a bit hard because I had to almost erase all the code to change the logic, but I loved the concept of state and the fact that it allows more flexibility and scalability to the project! 

I want to thank you a lot for your patience waiting for this PR. I really appreciate your time and all great reviews you and your colleagues are doing! ❤️ 🙏


src/main/kotlin/blackjack/model/Dealer.kt
```diff
@@ -0,0 +1,26 @@
...
+                is Started -> {
+                    state.run()
+                }
+                else -> state
+            }
+    }
+
+    fun shouldDraw(): Boolean {
+        return state is Running && state.hand.sumCards() < 17
+    }
```
Wordbe: When using the `is` operator, external branching logic is introduced to distinguish between the subclasses of the interface. Could we possibly implement this without using the `is` operator? 🙂

Also, I’d like the meaning of 17 to be clarified.

src/main/kotlin/blackjack/model/Deck.kt
```diff
@@ -0,0 +1,27 @@
...
+class Deck(val cards: MutableList<Card>) {
+    fun drawCard(): Card {
+        return cards.removeFirst()
+    }
+
+    companion object {
+        val cardPositionValues = listOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12)
+
+        fun generate(shuffled: Boolean): Deck {
+            val cardSuits = listOf(CardSuit.HEART, CardSuit.DIAMOND, CardSuit.CLUB, CardSuit.SPADE)
```
Wordbe: Just like you made `cardSuits` a local variable, I think it would be fine to encapsulate `cardPositionValues` as a local variable as well. 🙂

src/main/kotlin/blackjack/model/Hand.kt
```diff
@@ -0,0 +1,29 @@
...
+
+        return cardValues.sum()
+    }
+
+    companion object {
+        private const val ACE_LOW = 1
+        private const val ACE_HIGH = 11
+        private const val BLACKJACK_LIMIT = 21
+        private const val FACE_CARD_VALUE = 10
+        private val FACE_CARD_NUMBERS = setOf(11, 12, 13)
```
Wordbe, line 27: I’m not sure what 13 means. Could you explain it to me? 🤔

src/main/kotlin/blackjack/model/Participant.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    fun bet(value: Double) {
+        wallet.addToBet(value)
+    }
+
+    fun updateWalletWithEarningsRate(earningsRate: Double) {
+        wallet.updateWithEarningsRate(earningsRate)
+    }
+
+    fun calculateEarningsRate(p2: Participant): Double {
```
Wordbe: Shall we change `p2` to a more descriptive variable name? 🙂

src/main/kotlin/blackjack/state/Running.kt
```diff
@@ -0,0 +1,21 @@
...
+    override fun run(): State {
+        hand.addCard(deck.drawCard())
+        return evaluateNextState()
+    }
+
+    fun evaluateNextState(): State {
+        val sum = hand.sumCards()
+
+        return when {
+            sum == 21 -> this.finish()
```
Wordbe: Wouldn’t it be better to go with Blackjack? 🙂

src/main/kotlin/blackjack/state/Started.kt
```diff
@@ -0,0 +1,25 @@
...
+        }
+    }
+
+    override fun stay(): State {
+        return Stay(hand, deck)
+    }
+
+    override fun finish(): State {
+        return Finished(hand, deck)
+    }
```
Wordbe: I think we might not need `finish`. What do you think? 🤔

src/main/kotlin/blackjack/state/State.kt
```diff
@@ -0,0 +1,19 @@
...
+
+    fun run(): State
+
+    fun stay(): State
+
+    fun finish(): State
+
+    fun isBlackjack(): Boolean {
+        return hand.sumCards() == 21
+    }
```
Wordbe: This method is a public method that can be used externally, but it’s only being used internally. And as I understand it, *Started*, *Blackjack*, and *Finished* could all potentially be determined as Blackjack, which seems a bit ambiguous. How about making only the `Blackjack` state true, or removing it altogether if it’s not being used? 🤔

src/main/kotlin/blackjack/state/Started.kt
```diff
@@ -0,0 +1,25 @@
+package blackjack.state
+
+import blackjack.model.Deck
+import blackjack.model.Hand
+
+open class Started(override val hand: Hand, override val deck: Deck) : State {
```
Wordbe: The part I was confused about is that `Started` can transition to the next state `Running`, and `Running` is a subclass of `Started`. What about making `Started` and `Finished` abstract and expressing them only through their subclasses? I think that would make things clearer. 🙂

src/main/kotlin/blackjack/state/Stay.kt
```diff
@@ -0,0 +1,6 @@
+package blackjack.state
+
+import blackjack.model.Deck
+import blackjack.model.Hand
+
+class Stay(hand: Hand, deck: Deck) : Finished(hand, deck)
```
Wordbe: It seems that `Stay` and `Black` are just defined as states, but aren’t actually used anywhere. On the other hand, `Finished` is being used. We could consider utilizing this state when calculating the score. 🙂

src/main/kotlin/blackjack/view/InputView.kt
```diff
@@ -0,0 +1,40 @@
...
+                println(e.message)
+            } catch (e: NumberFormatException) {
+                println("$e .Please enter a valid number.")
+            }
+        }
+    }
+
+    fun askPlayerIfShouldHit(participant: Participant): Boolean {
+        println()
+        println("Would ${participant.name} like to draw another card? (y for yes, n for no)")
```
Wordbe, line 36: Try combining it into one line. 🙂

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fun printBetQuestion(player: Player) {
+        println(ASK_BET.format(player.name))
+    }
+
+    fun printPlayersIntroToTheirCards(players: List<Player>) {
+        val playersNames = players.map { it.name }
+        println()
+        println(FIRST_TURN_CARDS.format(playersNames.joinToString(",")))
```
Wordbe: Here as well.

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    fun formatCard(card: Card): String {
+        val cardValue = card.numberToCardDeckElements(card.number)
+        return "%s%s".format(cardValue, card.suit.value)
```
Wordbe: I recommend Kotlin's `string template` instead of this.

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,94 @@
...
+    fun printPlayersFirstTurnCards(players: List<Player>) {
+        players.forEach { player ->
+            val finalHand = formatHand(player.state.hand)
+            println(DISPLAY_HANDS.format(player.name, finalHand))
+        }
+    }
+
+    fun printDealerDrawsOneMoreCardMessage() {
+        println()
+        println(DEALER_DRAW_MESSAGE)
```
Wordbe: Same here.

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,94 @@
...
+    fun printCurrentDealerCards(dealer: Dealer) {
+        val cards = formatHand(dealer.state.hand)
+        println(DISPLAY_HANDS.format(dealer.name, cards))
+    }
+
+    fun printDealerFinalScore(dealer: Dealer) {
+        val cards = formatHand(dealer.state.hand)
+        val score = dealer.state.hand.sumCards()
+        println()
+        println(DISPLAY_FINAL_HAND.format(dealer.name, cards, score))
```
Wordbe: Also here!

src/main/kotlin/blackjack/view/OutputView.kt
```diff
@@ -0,0 +1,94 @@
...
+        players.forEach { player ->
+            val cards = formatHand(player.state.hand)
+            val score = player.state.hand.sumCards()
+            println(DISPLAY_FINAL_HAND.format(player.name, cards, score))
+        }
+    }
+
+    fun printFinalParticipantResult(participants: MutableMap<String, ParticipantWallet>) {
+        println()
+        println(DISPLAY_TITLE_FINAL_EARNINGS)
```
Wordbe: Here 🙂

Wordbe (changes requested):
Hello Gabi!

Thank you for carefully completing and submitting the code.
I’ve left some review comments, so please take a look and incorporate the ones you can.
You’re doing really well so far, so I hope you continue your hard work and finish strong.
Thank you! 🙂

src/main/kotlin/blackjack/model/Dealer.kt
knopgm: I added the shouldDraw behaviour to the state and override it inside the running state. 
It seams to make much more sense there. What do you think about it?

src/main/kotlin/blackjack/model/Deck.kt
knopgm: A lot of refactoring about this "small" review 😅, but totally worth it! 👍

src/main/kotlin/blackjack/model/Hand.kt
knopgm: I was confused about the values of Jack, Queen and King. 
Now, in a quick online check I saw that there is 13 card faces in a deck of cards. 
But there it was applied correctly -> if any card value is the same as one of those three numbers (11, 12, 13), then it will became the value of 10.

Let me know if it is clearer now. 🤞

src/main/kotlin/blackjack/state/Started.kt
knopgm: Hmm.. I'm sure it made sense when I used it, but now that I try to remove, everything is working fine without it. 😅 
Removed.

src/main/kotlin/blackjack/state/State.kt
knopgm: Hmm. You are right, the name isBlackjack() can be ambiguous, as it could be interpreted as a property of the State itself rather than a check on the hand's value. 😅 
My intention was to use this function as a reusable helper to simplify the logic within Started and other states, allowing them to make a decision about a state transition based on the hand's sum. It's a conditional check for a specific value, not a type identifier for the state.
To resolve this ambiguity and improve clarity, I will rename the function to canTransitionToBlackjack().

Let me know if like that it make more sense. 🤞

src/main/kotlin/blackjack/state/Started.kt
knopgm: Interesting 🤔, I refactored the states as you suggested. The logic is now split between "Initial" and "Running" states, which both derive from a new "Active" abstract class that means that the participant is not done yet.

src/main/kotlin/blackjack/state/Stay.kt
knopgm: I refactored so that the Finished states encapsulate their own scoring logic (via earningsAgainst), and ParticipantResult just delegates to them. That way the states are actually being used.
I hope it is making more sense now. 🤞

src/main/kotlin/blackjack/model/Dealer.kt
Wordbe: Yes, that sounds great. Thank you! 🙂

src/main/kotlin/blackjack/model/Deck.kt
Wordbe: When you change one thing, it often makes you want to change other related parts as well. 🤣
However, it seems the part I mentioned hasn’t been applied. If `cardNumbers` is only used in `generateCardsWithSuit`, I think it could be made a local variable of that method. 🙂

src/main/kotlin/blackjack/model/Hand.kt
Wordbe: In that case, I think it would be clearer to use *Jack*, *Queen*, and *King* instead of 11, 12, and 13. What do you think? 🙂

src/main/kotlin/blackjack/state/State.kt
Wordbe: One point to note is that methods defined in an interface are public and mainly used externally where the `State` is referenced. However, `canTransitionToBlackjack` is only used internally, specifically within the `Initial` subclass of `Active`. That means it isn’t needed in `Finished`. How about making `canTransitionToBlackjack` a private method inside the `Initial` object instead? 🤔

src/main/kotlin/blackjack/state/Started.kt
Wordbe: Great work!

src/main/kotlin/blackjack/state/Blackjack.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import blackjack.model.Deck
+import blackjack.model.Hand
+
+class Blackjack(hand: Hand, deck: Deck) : Finished(hand, deck) {
+    override fun earningsAgainst(other: Finished): Double {
+        return when (other) {
+            is Busted -> 1.5
+            is Blackjack -> 0.0
+            else -> 1.5 // wins against Stay
```
Wordbe: If you manage `Finished` as a sealed class, you can clearly distinguish cases with `is Stay`.
Also, if another state is later added under `Finished`, instead of relying on `else`, the compiler will raise an error, which helps ensure safer and more reliable code. Shall we give it a try? 🙂

Wordbe (changes requested):
Thank you for the quick feedback!
I’ve also left a bit more review, so please check it once more and apply the parts that are feasible.
Thanks again! 🙂

src/main/kotlin/blackjack/model/Deck.kt
knopgm: True!
I removed cardPositionValues entirely by switching to CardNumber.values(). That way, the card numbers are scoped locally to generateCardsWithSuit and we also avoid duplicating enum values.

src/main/kotlin/blackjack/model/Hand.kt
knopgm: Done! 😄

src/main/kotlin/blackjack/state/State.kt
knopgm: I see... You're right. I did as you said and it makes more sense now. 
Thank you for coming back to it. 🙏

src/main/kotlin/blackjack/state/Blackjack.kt
knopgm: Interesting!
I apply the sealed class and the else of Blackjack is gone. Now it is ensuring the exhaustive handling options for a Blackjack state.
Thank you for your patience going through my code again! 🙏

src/main/kotlin/blackjack/model/Deck.kt
Wordbe: I’d like `cardNumbers` to be inside the `generateCardsWithSuit` method. That’s all. 
It still hasn’t been applied yet. 🙂

src/main/kotlin/blackjack/model/Hand.kt
Wordbe: But it seems nothing has been changed 🙂

Wordbe (approved):
Hello Gabi!

I think it’s a good point to wrap up the Blackjack mission here.
Thanks to the effort you put into crafting the code, it has become very high-quality. Great job!
I hope this mission has helped you learn a lot like when to use inheritance and interfaces, how to implement polymorphism, and how to distribute responsibilities among objects.
You’ve really worked hard, and I wish you only the best in your journey ahead. Thank you for seeing this through to the end! 🙂

